PR #8, opened 2025-05-26
Replace versioneer with setuptools_scm

SBlokhuizen:
This replaces `versioneer` in favor of `setuptools_scm`. This also removes the following files, which I think are deprecated now as they are covered by `pyproject.toml`, but please correct me if I am wrong:
- MANIFEST.in
- setup.cfg


maarten-ic (approved):
Looks good to me!

PR #9, opened 2025-05-27
Remove outdated VTK GGD Writer plugin

maarten-ic:
VTK to GGD conversion has not worked for a long time and we have decided that we will not work on this functionality. This commit removes the plugin to prevent any confusion for users.


SBlokhuizen:
I am fine with removing this for the reasons you stated, though I think then we should also remove the unused implementations of this plugin. For example, I believe `write_geom.py`, `write_ps.py`, `representables.py` are only used for the vtkggdwriter, and some other objects, for example the `FauxIndexMap` in `imas_paraview.util`.

SBlokhuizen (approved):

PR #10, opened 2025-05-28
Rename LoS Reader to Line of Sight Reader

SBlokhuizen:
This changes the name of the `LoS Reader` to the unabbreviated `Line of Sight Reader`. This makes the purpose of the reader more clear. The documentation was also updated to reflect this change.

Screenshot (5)


maarten-ic (approved):
💯

PR #17, opened 2025-06-06
Assume that IDSAndOccurrence receives a valid value from ParaView

### LINKED ISSUE #16: Legacy URI Input in ParaView Plugin Fails to Load Saved State Properly
The legacy URI input in the ParaView plugin fails to load correctly from a saved state. However, using the URI string input seems to work. 

On SDCC:
- `/home/ITER/blokhus/paraview_states/test_paulo.pvsm`  -  This inputs the URI of the Line of Sight reader using the legacy input, this has the issue of not loading properly from a saved state.
- `/home/ITER/blokhus/paraview_states/test_paulo_non_legacy_los.pvsm`  - This one inputs the URI of the Line of Sight reader using the URI string, this one has no issue.

maarten-ic:
Skipping the check fixes #16: when restoring ParaView state, the IDS and Occurrence parameter was restored before all URI parameters. Skipping this check has fixed the issue.

I didn't observe any adverse side effects while testing, but we should keep an eye out for any.

### DIFF
--- a/imas_paraview/plugins/base_class.py
+++ b/imas_paraview/plugins/base_class.py
@@ -281,7 +281,7 @@ class GGDVTKPluginBase(VTKPythonAlgorithmBase, ABC):
         selecting a URI to load the Data Entry), or when the loaded Data Entry contains
         no IDSs supported by this plugin.
         """
-        if value not in self._ids_list or value == "<Select IDS>":
+        if value == "<Select IDS>":
             self._selectable = []
             self._selected = []
             value = ""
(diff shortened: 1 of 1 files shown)


SBlokhuizen (approved):
LGTM!

SBlokhuizen (changes requested):
On further testing, it seems to not work for all readers yet. Take a look at the following saved state on SDCC:
`/home/ITER/blokhus/paraview_states/test_paulo.pvsm`

The GGDReader1 (which uses the legacy input) does work, but the Line of Sight reader and the wall limiter reader still do not load properly yet.

imas_paraview/plugins/base_class.py
```diff
@@ -281,7 +281,7 @@ def P10_SetIDSAndOccurrence(self, value):
         selecting a URI to load the Data Entry), or when the loaded Data Entry contains
         no IDSs supported by this plugin.
         """
-        if value == "<Select IDS>":
```
maarten-ic: Looks like this change was reverted?

SBlokhuizen (approved):

paulotex (approved):
Tested with my own saved state and everything is working !

PR #20, opened 2025-06-12
Bugfix/18 description ggd

### LINKED ISSUE #18: Issue loading description GGD data
There is an issue when loading the `wall` IDS of JOREK grids. For example:
```
imas:hdf5?user=public;pulse=112111;run=2;database=ITER_DISRUPTIONS;version=4
```
In this IDS the `description_ggd[0]` refers to the innter ITER vacuum vessel, whilst `description_ggd[1]` stores the first wall surface (as explained in more detail [here](https://confluence.iter.org/display/IMP/The+JOREK+disruption+cases)).

We assume that indices of description Grid GGDs behave similar to the Grid GGD, and thus are time based. This leads to the issue, where if we select time step 1 in ParaView, we see the first wall surface, while if we load time step 0 or >1, it loads the inner vacuum vessel (as it defaults to the first time step if the time step cannot be found).

Offending code: https://github.com/iterorganization/IMAS-ParaView/blob/develop/imas_paraview/util.py#L70

maarten-ic:
TODO:

- [x] Update documentation

Fixes #18:

- Update get_grid_ggd logic to use the time index only for time-dependent arrays of structures
- Add logic to allow selecting different "parent" arrays of structures. For example, now it's possible to see `description_ggd[0]` from the wall IDS or `description_ggd[1]` from the wall IDS in the ``imas:hdf5?user=public;pulse=112111;run=2;database=ITER_DISRUPTIONS;version=4`` data entry.

image
image


SBlokhuizen:
Nice work, tested it out with some IDSs and it works well! LGTM

SBlokhuizen (approved):

PR #24, opened 2025-06-17
Dynamically load GGD arrays

SBlokhuizen:
The `phi` [GGD quantity](https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/wall.html#wall-description_ggd-ggd-j_total-phi) was renamed from `toroidal` in DD version 3.42.0. As the GGD vectors are not dynamically searched for, `phi` was not being loaded. 

The loading of GGD vectors was changed such that it loads all filled FLT 1D arrays that it encounters instead, so that this does not happen in the future.

### DIFF
--- a/imas_paraview/io/read_ps.py
+++ b/imas_paraview/io/read_ps.py
@@ -320,22 +320,13 @@ class PlasmaStateReader:
 
         # Only add the components that have data:
         components = dict()  # name and values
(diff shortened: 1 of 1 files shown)


paulotex (approved):

maarten-ic (approved):
🚀

PR #25, opened 2025-06-18
Export cells in 2D profiles reader

SBlokhuizen:
Instead of exporting separate points in the unstructured grid for the 2D profile reader, export quadrilateral cells.|

image

### DIFF
--- a/imas_paraview/plugins/profiles_2d.py
+++ b/imas_paraview/plugins/profiles_2d.py
@@ -258,28 +258,26 @@ class Profiles2DReader(GGDVTKPluginBase, is_time_dependent=True):
         return vtk_scalars
 
     def _create_ugrid(self, vtk_points, vtk_scalars):
-        """Create a vtkUnstructuredGrid of the given profile.
+        """Create a vtkUnstructuredGrid using VTK_QUAD cells from the structured 2D grid."""
 
-        Args:
-            profile: The profile to create a ugrid for.
-
-        Returns:
-            The created unstructured grid.
-        """
-
-        ugrid = vtk.vtkUnstructuredGrid()
-        ugrid.GetPointData().SetScalars(vtk_scalars)
-
-        num_points = vtk_points.GetNumberOfPoints()
-        cells = vtk.vtkCellArray()
-
-        for i in range(num_points):
-            vertex = vtk.vtkVertex()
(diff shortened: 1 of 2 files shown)


imas_paraview/plugins/profiles_2d.py
```diff
@@ -258,28 +258,32 @@ def _create_vtkscalars(self, profile):
...
-        ugrid.GetPointData().SetScalars(vtk_scalars)
-        ugrid.SetCells(vtk.VTK_VERTEX, cells)
+        for i in range(n_rows - 1):
+            for j in range(n_cols - 1):
+                quad = vtk.vtkQuad()
+                quad.GetPointIds().SetId(0, get_index(i, j))
+                quad.GetPointIds().SetId(1, get_index(i, j + 1))
+                quad.GetPointIds().SetId(2, get_index(i + 1, j + 1))
+                quad.GetPointIds().SetId(3, get_index(i + 1, j))
+                cells.InsertNextCell(quad)
```
maarten-ic: This is actually pretty slow for larger grids (256 x 512 profiles_2d would already take a second on my laptop).

Can we do something smarter with numpy, like the following (not tested though 😉 )?
```suggestion
        quad = np.array([[0, 1, n_cols, n_cols + 1]])
        all_quads = quad + np.arange((n_cols - 1) * (n_rows - 1))[:, None]
        cells.SetCells(all_quads.shape[0], numpy_to_vtk(all_quads, deep=True, array_type=vtk.VTK_ID_TYPE))
```

pretty similar to how [the jorek reader populates the bezier cells](https://github.com/iterorganization/IMAS-ParaView/blob/develop/imas_paraview/io/read_jorek.py#L256)

maarten-ic (commented):
Nice! One comment on the performance, can you have a look at that?

imas_paraview/plugins/profiles_2d.py
SBlokhuizen: Good call! Your implementation gave me segfaults upon loading, but I did manage to get it working. Loading my test grid I have an improvement of ~40x (0.8s vs 0.01s). Let me know what you think!

maarten-ic: Very nice, thanks! 😄

maarten-ic (approved):

PR #27, opened 2025-06-24
Feature/remember selected attribute arrays

maarten-ic:
This PR introduces a slight refactor of the array selection logic, but more importantly: the IMAS-ParaView plugins now remember which attribute arrays are selected when switching URI or IDS/occurrence.

Example video below:

1. Data loaded from one URI 
2. Switch to a different URI
3. Data array (Electrons density) remains selected

https://github.com/user-attachments/assets/0b367434-4e1e-4551-84c2-03bf2ac93645


SBlokhuizen (commented):
Nice! Looks good! Now that it stores the attribute arrays upon changing URI, we don't have to store and reload them in the batch scripts. Could you update the docs [here](https://github.com/iterorganization/IMAS-ParaView/blob/develop/docs/source/training_batch.rst?plain=1#L39), and also check to make sure the script still works?

maarten-ic:
> Nice! Looks good! Now that it stores the attribute arrays upon changing URI, we don't have to store and reload them in the batch scripts. Could you update the docs [here](https://github.com/iterorganization/IMAS-ParaView/blob/develop/docs/source/training_batch.rst?plain=1#L39), and also check to make sure the script still works?

Good point! Will do

SBlokhuizen (approved):

PR #28, opened 2025-06-27
Add support for heterogeneous time in GGD readers (incl. JOREK)

maarten-ic:
Tested successfully with heterogeneous time JOREK disruption dataset: `imas:hdf5?user=public;pulse=113111;run=3;database=ITER_DISRUPTIONS;version=4`

### DIFF
--- a/imas_paraview/util.py
+++ b/imas_paraview/util.py
@@ -55,7 +55,7 @@ def get_ggd_path(ids_metadata) -> Optional[str]:
     return None
 
 
-def get_grid_ggd(ids, ggd_idx=0, parent_idx=0):
+def get_grid_ggd(ids, time=0, parent_idx=0):
     """Finds and returns the grid_ggd within IDS at the time index ggd_idx. If the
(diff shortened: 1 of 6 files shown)


imas_paraview/util.py
```diff
@@ -55,7 +55,7 @@ def get_ggd_path(ids_metadata) -> Optional[str]:
     return None
 
 
-def get_grid_ggd(ids, ggd_idx=0, parent_idx=0):
+def get_grid_ggd(ids, time=0, parent_idx=0):
```
SBlokhuizen, line 58: Can you update the docstring as well?

SBlokhuizen (commented):
Nice work! One minor comment :)

SBlokhuizen (approved):

imas_paraview/util.py
SBlokhuizen: As a sidenote, perhaps also turn warnings (like we did with the waveform editor) into errors for IMAS-Paraview. That should catch such things in the future

SBlokhuizen (approved):

PR #30, opened 2025-07-07
Format units with unicode superscript characters

maarten-ic:
### DO NOT MERGE

Note: paraview issue https://gitlab.kitware.com/paraview/paraview/-/issues/23016 prevents this from displaying properly in the Renderview legends.


paulotex:
> ### DO NOT MERGE
> 
> Note: paraview issue https://gitlab.kitware.com/paraview/paraview/-/issues/23016 prevents this from displaying properly in the Renderview legends.

Note (for future reference): the paraview issue seems to be related to the default fonts lacking complete unicode support:

- If we specify a unicode font, all unicode chars are rendered properly
- If we introduce LaTeX markup, all unicode chars are rendered properly

PR #31, opened 2025-07-10
Add tests with real IDS data

SBlokhuizen:
This PR introduces updated and expanded test coverage using real IDS data stored in NetCDF format. It includes:

- Test for the 1D profile reader using JINTRAC dataset.
- New tests for loading GGDs using SOLPS and JOREK datasets.
- Existing test datasets are converted to NetCDF and stored in a local directory on SDCC: `/home/ITER/blokhus/public/imas_paraview_tests`

All tests been run and pass when executed on SDCC.

Next Steps:

- [x] Request approval for publication of test datasets.
- [x] The IDSs are fetched from local SDCC directories, these should be fetched elsewhere, and ideally cached in the CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,12 +108,14 @@ jobs:
         key: ${{ steps.cache-paraview.outputs.cache-primary-key }}
         path: paraview/
 
-    - name: Install dependencies
-      run: |
-        # Install dependencies required for running paraview gui in xfvb-run:
-        sudo apt update
-        sudo apt install -y libqt5gui5-gles
+    - name: Cache and Install Dependencies
+      uses: awalsh128/cache-apt-pkgs-action@latest
+      with:
+        packages: libqt5gui5-gles
+        version: 1.0
 
+    - name: Install Python dependencies
+      run: |
         # Install python dependencies
         python -m venv venv
         . venv/bin/activate
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ name: CI
 on: push
 
 env:
-  # Zenodo record ID for test datasets
+  # Zenodo record IDs for test datasets
   ZENODO_MD_ID: 17113713
   ZENODO_SCENARIOS_ID: 17062700
 
@@ -70,14 +70,14 @@ jobs:
         key=$(echo "${urls[*]}" | sha256sum | cut -d ' ' -f1)
         echo "key=${key}" >> $GITHUB_OUTPUT
 
-    - name: Restore Zenodo datasets from cache
-      id: cache-datasets-restore
-      uses: actions/cache/restore@v4
+    - name: Cache Zenodo datasets
+      id: cache-datasets
+      uses: actions/cache@v4
       with:
         path: data
         key: ${{ steps.datasets.outputs.key }}
 
-    - if: ${{ steps.cache-datasets-restore.outputs.cache-hit != 'true' }}
+    - if: ${{ steps.cache-datasets.outputs.cache-hit != 'true' }}
       name: Download datasets if not cached
       run: |
         mkdir -p data
@@ -85,13 +85,6 @@ jobs:
           echo "Downloading $(basename $url) ..."
           wget -P data/ "$url"
         done
-
-    - if: ${{ steps.cache-datasets-restore.outputs.cache-hit != 'true' }}
-      name: Save Zenodo datasets to cache
-      uses: actions/cache/save@v4
-      with:
-        path: data
-        key: ${{ steps.datasets.outputs.key }}      
         
     - name: Cache paraview
       id: cache-paraview
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,11 +2,6 @@ name: CI
 
 on: push
 
-env:
-  # Zenodo record IDs for test datasets
-  ZENODO_MD_ID: 17113713
-  ZENODO_SCENARIOS_ID: 17062700
-
 jobs:
   linting:
     runs-on: ubuntu-latest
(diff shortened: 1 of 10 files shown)


imas_paraview/tests/test_beam.py
```diff
@@ -1,22 +1,16 @@
...
 def test_load_beam():
     """Test if limiters are loaded in the VTK Multiblock Dataset."""
     reader = BeamReader()
     with DBEntry(
-        (
-            "imas:hdf5?path=/work/imas/shared/imasdb/ITER_MACHINE_DESCRIPTION/"
-            "3/120000/1304/"
-        ),
-        "r",
+        "/home/ITER/blokhus/public/imas_paraview_tests/iter_md-120000-1304.nc", "r"
```
maarten-ic: Perhaps you could make the "base folder" where the test data is stored a fixture (in conftest.py).

Once the datasets are published, we would only need to update one fixture instead of every test file.

maarten-ic (commented):
Looks good! One small comment:

maarten-ic (approved):

olivhoenen:
Open access data is finally here: 
- machine description at https://doi.org/10.5281/zenodo.15525525
- scenarios at https://doi.org/10.5281/zenodo.17062700

SimonPinches (approved):
lgtm

paulotex (approved):

paulotex (commented):
The CI tests are failing. It seems some files are not found...

` E   FileNotFoundError: [Errno 2] No such file or directory: 'data/iter_sce_110004-1.nc'`

olivhoenen:
> The CI tests are failing. It seems some files are not found...
> 
> ` E FileNotFoundError: [Errno 2] No such file or directory: 'data/iter_sce_110004-1.nc'`

As discussed with @SBlokhuizen on slack, this pulse was not added to the list of data publish open access on zenodo, but maybe you can use the equilibrium from [iter_scenario_53298_seq1_DD3.nc](https://zenodo.org/records/17062700/files/iter_scenario_53298_seq1_DD3.nc?download=1) instead?

SBlokhuizen:
@paulotex There were two issues left to resolve that I discussed yesterday with Olivier, the first one you mentioned is resolved in the last commit. The second issue is related to the magnetics IDS used in the test for the position reader. 

The magnetics IDS used for testing was created with DD 3.40.1. In this version the magnetic probe IDS still contained [a unique identifier](https://imas-data-dictionary.readthedocs.io/en/3.42.2/generated/ids/magnetics.html#magnetics-b_field_pol_probe-identifier) field, as well as [a name](https://imas-data-dictionary.readthedocs.io/en/3.42.2/generated/ids/magnetics.html#magnetics-b_field_pol_probe-name). It seems that in DD 4.0.0 this has been replaced with a [single unique name](https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/magnetics.html#magnetics-b_field_phi_probe-name). In the conversion from DD3 to DD4, the name is simply copied and the identifier field is removed, which means that the devices are not unique identifiable anymore, which goes against the DD definition (For example, `ids.b_field_pol_probe[0 - 177].name = '55.A3 Tangential Coils (Outer)'`). 

IMAS-ParaView requires the objects to be uniquely identifiable by their name + optionally an identifier. As a remedy, we could label duplicate entries with a marker, so that duplicate names are supported, for example:

- 55.A3 Tangential Coils (Outer)
- 55.A3 Tangential Coils (Outer) (1)
- 55.A3 Tangential Coils (Outer) (2)

What do you think?

SBlokhuizen:
@paulotex, @olivhoenen I just tested the last failing test (`imas_paraview/tests/test_position.py::test_load_position_magnetics`) using the magnetics IDS converted to DD version 4.0.0 using [Deepak's conversion fix](https://github.com/iterorganization/IMAS-Python/pull/68). This solves the issue I described in previous message, and the test succeeds. 

@olivhoenen Could you update the magnetics machine description IDS on Zenodo to one converted with this fix? Thanks!

olivhoenen:
Great, I will do that asap.

olivhoenen:
Should be done, check the v2 entry here: https://doi.org/10.5281/zenodo.17113713

.github/workflows/ci.yml
```diff
@@ -45,6 +50,49 @@ jobs:
...
+          https://zenodo.org/record/${ZENODO_MD_ID}/files/iter_md_wall_116000_5.nc
+          https://zenodo.org/record/${ZENODO_MD_ID}/files/iter_md_wall_116100_2001.nc
+        )
+        echo "files=${urls[*]}" >> $GITHUB_OUTPUT
+
+        # Generate a hash key from the URLs list, this is used for caching
+        key=$(echo "${urls[*]}" | sha256sum | cut -d ' ' -f1)
+        echo "key=${key}" >> $GITHUB_OUTPUT
+
+    - name: Restore Zenodo datasets from cache
```
DaanVanVugt: if you use the 'complete' cache action you don't need to save explicitly in a second step

.github/workflows/ci.yml
```diff
@@ -45,6 +50,49 @@ jobs:
...
+    - if: ${{ steps.cache-datasets-restore.outputs.cache-hit != 'true' }}
+      name: Download datasets if not cached
+      run: |
+        mkdir -p data
+        for url in ${{ steps.datasets.outputs.files }}; do
+          echo "Downloading $(basename $url) ..."
+          wget -P data/ "$url"
+        done
+
+    - if: ${{ steps.cache-datasets-restore.outputs.cache-hit != 'true' }}
```
DaanVanVugt: so this one could go away

.github/workflows/ci.yml
```diff
@@ -70,6 +118,7 @@ jobs:
     - name: Install dependencies
       run: |
         # Install dependencies required for running paraview gui in xfvb-run:
+        sudo apt update
```
DaanVanVugt: can use https://github.com/marketplace/actions/cache-apt-packages instead?

.github/workflows/ci.yml
```diff
@@ -45,6 +50,49 @@ jobs:
         python-version: '3.10'
         cache: 'pip'  # caching pip dependencies
 
+    - name: Define dataset list and cache key
+      id: datasets
+      run: |
+        urls=(
+          https://zenodo.org/record/${ZENODO_SCENARIOS_ID}/files/iter_disruption_113112_1.nc
```
DaanVanVugt: the filenames are not obvious per record ID. Perhaps even just list the filenames and construct the URLs from that? On the other hand this is not so critical.

DaanVanVugt (approved):

.github/workflows/ci.yml
SBlokhuizen: As discussed, replaced the environment variables with a simple list of URLs for simplicity.

olivhoenen (approved):
All tests are passing now, thanks @SBlokhuizen !

PR #39, opened 2025-11-26
Add support for more 3D volume types

### LINKED ISSUE #34: Visualizing 3D elements
I'm trying to visualize a wall with 3D elements. 
However, the grid in the description_ggd works only for 2D elements.
Instead, I defined grid_subset, as they seem to take into account 3D objects, but this results in the 
error vtkDataAssembly.cxx:530    ERR| vtkDataAssembly (0x9e5f4e0): Invalid name specified 

I'm not sure if this is a bug in IMAS-Paraview or I defined the ids wrong.
Do you have an example of an IDS with 3D elements, so I can double-check?
I have been using: imas:hdf5?user=schwarn;pulse=100000;run=2;database=CARIDDI;version=4

SBlokhuizen:
Proposed solution for #34. Instead of always returning a VTK polyhedron, return the appropriate VTK volume based on how many points are provided. A few of the following shapes are covered:

https://examples.vtk.org/site/Cxx/GeometricObjects/Cell3DDemonstration/
image

Tested and working with the following dataset (mentioned in #34):
`imas:hdf5?user=schwarn;pulse=100000;run=2;database=CARIDDI;version=4`

image


maarten-ic (approved):
LGTM

N-Schwarz:
Thank you!, this is working now and it also displays the values from the ggd. 
However, the components 1-3 are missing from the vtk. 
And also only the subsets are displayed and not the complete mesh (the data from grid_ggd[0].space[0]).
For me, this is fine because we are more interested in the components. 
But maybe other applications do not define subsets of the grid.

SBlokhuizen:
Hi @N-Schwarz and @paulotex, I have taken a look at it, and there seemed to be an issue with the conversion logic.
I created a fix, and it seems to be working with the following URIs:
- `imas:hdf5?user=schwarn;pulse=100000;run=2;database=CARIDDI;version=4`
- `imas:hdf5?user=public;pulse=116100;run=1001;database=ITER_MD;version=3`

Could you take a look at it to see if this works for you now?

Before merging this PR, the following still has to be done:
- [x] Test with other IDSs if no regressions are introduced
- [x] Create new test data and write tests to prevent issues in the future

N-Schwarz:
Sorry for the late reply. I do not have access to the ITER cluster or the data at the moment. But the figure looks good to me and the version before already worked for my needs. Thanks a lot!

imas_paraview/tests/fill_ggd.py
```diff
@@ -86,62 +91,83 @@ def fill_NxN_grid(grid_ggd, N, create_3d_grid=False):
...
+def build_grid(N, space, create_3d_grid, create_volumes):
     """Builds the vertices, edges and faces for an N x N grid.
 
     Args:
         N: Size of the grid
         space: Space AoS of the GGD grid
         create_3d_grid: If True, create a 2D grid in 3D space (in the X-Z plane at
             Y = 0). If False, create a 2D grid in the X-Y plane.
+        create_volumes: If True, extrude the 2D grid by in the Z direction to
+            create a layer of cube volume cells. Requires create_3d_grid=True.
```
Yannicked, line 124: Add a check for it if required

Yannicked (approved):

PR #40, opened 2026-01-13
Disable mdsplus/hdf5 integration tests

SBlokhuizen:
There are some issues with the MDSPlus and HDF5 integration tests using the open-source version of imas-core.  Let's skip them for now.

### DIFF
--- a/imas_paraview/tests/conftest.py
+++ b/imas_paraview/tests/conftest.py
@@ -35,6 +35,21 @@ def set_environment():
     print(f"PYTHONPATH={os.environ['PYTHONPATH']}")
 
 
+def pytest_addoption(parser):
+    parser.addoption(
+        "--skip-hdf5",
+        action="store_true",
+        default=False,
+        help="Skip integration tests for HDF5 datasets",
+    )
+    parser.addoption(
+        "--skip-mdsplus",
+        action="store_true",
+        default=False,
+        help="Skip integration tests for MDSplus datasets",
+    )
+
+
 def pytest_sessionstart(session):
     """Set up the environment variables and generate test data to
     perform the integration tests on."""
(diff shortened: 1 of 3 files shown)


imas_paraview/tests/conftest.py
```diff
@@ -56,17 +56,18 @@ def pytest_sessionstart(session):
...
+    # with imas.DBEntry(
+    #     "imas:mdsplus?path=mdsplus_testdb", "w", dd_version=DD_VERSION
+    # ) as dbentry:
+    #     dbentry.put(ids)
+    #
+    # # Write test file as HDF5
+    # with imas.DBEntry(
+    #     "imas:hdf5?path=hdf5_testdb", "w", dd_version=DD_VERSION
+    # ) as dbentry:
+    #     dbentry.put(ids)
```
maarten-ic: The HDF5 backend should work fine, why is it disabled as well?

imas_paraview/tests/conftest.py
```diff
@@ -56,17 +56,18 @@ def pytest_sessionstart(session):
...
-        "imas:mdsplus?path=mdsplus_testdb", "w", dd_version=DD_VERSION
-    ) as dbentry:
-        dbentry.put(ids)
-
-    # Write test file as HDF5
-    with imas.DBEntry(
-        "imas:hdf5?path=hdf5_testdb", "w", dd_version=DD_VERSION
-    ) as dbentry:
-        dbentry.put(ids)
+    # TODO: Enable hdf5/mdsplus integration tests using open-source imas-core
```
maarten-ic: AFAIK there are no plans to enable MDSplus in the Python wheels of imas-core

imas_paraview/tests/conftest.py
paulotex: I am wondering the same...

SBlokhuizen: Sorry, I should have been more clear!

Currently, there is an issue with MDSplus backend when using the public `imas-core` package. As I have heard from Maarten, there are no plans to resolve this in the near future. 

As for the HDF5 integration tests, there is a discrepancy between the HDF5 library version used in ParaView binary that we use in the CI, and the one in imas-core causing conflicts. 

I have now created two flags which you can pass to pytest:
- `--skip-mdsplus` - This skips the integration tests using **MDSplus** IDS data.
- `--skip-hdf5` - This skips the integration tests using **HDF5** IDS data.

I have turned both flags on in the CI for now, such that the MDSplus and HDF5 integration tests are skipped and the CI passes. By default the these integration tests will still run (for example when running `pytest` on SDCC).

Note: The MDSplus and HDF5 integration tests were not covered in the CI before, so we are not losing coverage here!

maarten-ic (approved):

PR #41, opened 2026-01-14
Reader for pf_active and pf_passive structures

SBlokhuizen:
This introduces a new ParaView plugin: the `PF Reader` (Feel free to suggest a better name!)

This reader allows you to visualize the coil geometry from the pf_active IDS, as well as the loop geometry from the pf_passive IDS. This has been tested on the following URIs, which contain the following geometry types:

| IDS        | URI                                                          | Geometries              |
|------------|--------------------------------------------------------------|-------------------------|
| pf_active  | imas:hdf5?path=/work/imas/shared/imasdb/ITER_MD/3/111001/104 | `rectangle`             |
| pf_active  | imas:hdf5?path=/work/imas/shared/imasdb/ITER_MD/3/111001/204 | `rectangle`, `annulus`  |
| pf_passive | imas:hdf5?path=/work/imas/shared/imasdb/ITER_MD/3/124002/3   | `thick line`            |
| pf_passive | imas:hdf5?path=/work/imas/shared/imasdb/ITER_MD/3/124001/3   | `thick line`            |
| pf_passive | imas:hdf5?path=/work/imas/shared/imasdb/ITER_MD/3/116001/3   | `oblique`               |
| pf_passive | imas:hdf5?path=/work/imas/shared/imasdb/ITER_MD/3/115005/3   | `rectangle`,`oblique`   |
| pf_passive | imas:hdf5?path=/work/imas/shared/imasdb/ITER_MD/3/115004/6   | `annulus`, `thick line` |

A note on the two pf_active IDSs in the table above; occurrence=1 of these IDSs contains rectangle geometries where `width = height = 0`, these cannot be visualized with the default surface representation in ParaView, so you must select the  `Points` geometry representation.

I have not been able to find machine description IDSs containing geometry types for `outline` or `arcs_of_circle`, which is why these have not been implemented. If there are IDSs which do contain these geometry types, we can add them.

The CI fails due to the issues mentioned in #40, once this is merged the CI should pass.


maarten-ic:
> I have not been able to find machine description IDSs containing geometry types for outline or arcs_of_circle, which is why these have not been implemented. If there are IDSs which do contain these geometry types, we can add them.

It shouldn't be too difficult to generate a test IDS which covers all geometry types, should it?

SBlokhuizen:
> > I have not been able to find machine description IDSs containing geometry types for outline or arcs_of_circle, which is why these have not been implemented. If there are IDSs which do contain these geometry types, we can add them.
> 
> It shouldn't be too difficult to generate a test IDS which covers all geometry types, should it?

I have now added support for the `outline` and `arcs_of_circle` geometry types, including a test case for each.

SBlokhuizen:
Something to consider; the following IDSs contain the same geometry types as used in `pf_active` and `pf_passive`: 
- [ferritic](https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/ferritic.html#ferritic-object-axisymmetric)
- [ic_antennas](https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/ic_antennas.html#ic_antennas-antenna-module-strap-geometry)
- [iron_core](https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/iron_core.html#iron_core-segment-geometry)

We should also be able to visualize these geometries using this reader, although I have not been able to find any machine description IDSs which actually fill this. Perhaps we should rename this then to something more general e.g. `Axisymmmetric Geometry Reader`?

maarten-ic:
> Something to consider; the following IDSs contain the same geometry types as used in `pf_active` and `pf_passive`:
> 
> * [ferritic](https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/ferritic.html#ferritic-object-axisymmetric)
> * [ic_antennas](https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/ic_antennas.html#ic_antennas-antenna-module-strap-geometry)
> * [iron_core](https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/iron_core.html#iron_core-segment-geometry)
> 
> We should also be able to visualize these geometries using this reader, although I have not been able to find any machine description IDSs which actually fill this. Perhaps we should rename this then to something more general e.g. `Axisymmmetric Geometry Reader`?

Would be nice if we can plot those indeed! `iron_core` is used for WEST and should be available in some of the WEST datasets on SDCC, perhaps you could test with that?
As for the naming, `Axisymmmetric Geometry Reader` would not be immediately obvious to me when you'd need to choose this (e.g. wall 2D is also axisymmetric but cannot be read by this one). @paulotex do you have a preference or suggestion?

docs/source/pf_reader.rst
```diff
@@ -0,0 +1,34 @@
+.. _`using the PF Reader`:
+
+PF Reader
```
maarten-ic: Could we add an image here based on public data?

imas_paraview/plugins/pf.py
```diff
@@ -0,0 +1,273 @@
...
+        self.selectable_map = {}
+
+        if self._ids.metadata.name == "pf_active":
+            self._load_ids_quantities(self._ids.coil)
+        elif self._ids.metadata.name == "pf_passive":
+            self._load_ids_quantities(self._ids.loop, default_name="Loop")
+        else:
+            raise NotImplementedError(f"Unable to load {self._ids.metadata.name}.")
+
+    def _load_ids_quantities(self, ids_quantity, default_name="Coil"):
```
maarten-ic: Please don't provide a default argument if that's not generally applicable.

```suggestion
    def _load_ids_quantities(self, ids_quantity, default_name):
```

and then on line 37: `self._load_ids_quantities(self._ids.coil, default_name="Coil")`

imas_paraview/plugins/pf.py
```diff
@@ -0,0 +1,273 @@
...
+            default_name: Default name of the quantity.
+        """
+        for i, quantity in enumerate(ids_quantity):
+            quantity_name = quantity.name
+            if not quantity_name:
+                quantity_name = f"{default_name} {i}"
+                logger.warning(
+                    f"{default_name} without name found. "
+                    f"Renaming it to {quantity_name!r}"
+                )
```
maarten-ic: Better to use logger formatting instead, see https://docs.astral.sh/ruff/rules/logging-f-string/

```suggestion
                logger.warning(
                    "%s without name found. Renaming it to %r",
                    default_name,
                    quantity_name,
                )
```

maarten-ic: Can you enable these rules for `ruff` in `pyproject.toml`? Fine to do in a separate PR if that raises issues in code not touched in this PR :)

imas_paraview/plugins/pf.py
```diff
@@ -0,0 +1,273 @@
...
+                block_id += 1
+
+            logger.info(
+                f"Loaded {quantity_name!r} with {len(quantity.element)} element(s)."
+            )
+
+    def _create_outline(self, outline):
+        """Create vtkPolyData object from outline geometry"""
+        r, z = outline.r, outline.z
+        points = [(r_i, 0.0, z_i) for r_i, z_i in zip(r, z)]
```
maarten-ic: With strict=True an exception is raised when r and z have a different length
```suggestion
        points = [(r_i, 0.0, z_i) for r_i, z_i in zip(r, z, strict=True)]
```

imas_paraview/plugins/pf.py
```diff
@@ -0,0 +1,273 @@
...
+
+        points = [
+            (r0, 0.0, z0),
+            (r0 + dr_alpha, 0.0, z0 + dz_alpha),
+            (r0 + dr_alpha + dr_beta, 0.0, z0 + dz_alpha + dz_beta),
+            (r0 + dr_beta, 0.0, z0 + dz_beta),
+        ]
+        return points_to_vtkpoly(points, is_closed=True)
+
+    def _create_arcs_of_circle(self, arcs_of_circle, resolution=10):
```
maarten-ic: Can you make the resolution a parameters that users can tweak in the plugin?

imas_paraview/plugins/pf.py
```diff
@@ -0,0 +1,273 @@
...
+        outer_ids = []
+        inner_ids = []
+        pts = vtkPoints()
+
+        for i in range(resolution):
+            theta = 2.0 * np.pi * i / resolution
+            c = np.cos(theta)
+            s = np.sin(theta)
+            outer_ids.append(pts.InsertNextPoint(r0 + r_out * c, 0.0, z0 + r_out * s))
+            inner_ids.append(pts.InsertNextPoint(r0 + r_in * c, 0.0, z0 + r_in * s))
```
maarten-ic: It feels weird to do this as an explicit for-loop. Why not use numpy to do this vectorized and convert to VTK objects with https://docs.vtk.org/en/latest/api/python/vtkmodules/vtkmodules.util.numpy_support.html?

Same applies to the arcs-of-circle logic, I guess.

imas_paraview/tests/test_pf.py
```diff
@@ -0,0 +1,170 @@
...
+    else:
+        node = ids.loop
+    node.resize(1)
+    node[0].element.resize(1)
+    return node[0].element[0].geometry
+
+
+def get_points(poly):
+    pts = poly.GetPoints()
+    return [pts.GetPoint(i) for i in range(pts.GetNumberOfPoints())]
```
maarten-ic: Can't you use https://docs.vtk.org/en/latest/api/python/vtkmodules/vtkmodules.util.numpy_support.html#vtkmodules.util.numpy_support.vtk_to_numpy ?

imas_paraview/tests/test_pf.py
```diff
@@ -0,0 +1,170 @@
...
+    points = get_points(result)
+    assert len(points) == 4
+    expected_pts = np.array(
+        [
+            [1.0, 0.0, 2.0],
+            [1.0 + 1.5 * np.sqrt(3), 0.0, 3.5],
+            [3.0 + 1.5 * np.sqrt(3), 0.0, 3.5 + 2.0 * np.sqrt(3)],
+            [3.0, 0.0, 2.0 + 2.0 * np.sqrt(3)],
+        ]
+    )
```
maarten-ic: N.B. you don't have to explicitly create an array here (np.allclose will happily work with nested lists). Can save some effort and lines:


```suggestion
    expected_pts = [
        [1.0, 0.0, 2.0],
        [1.0 + 1.5 * np.sqrt(3), 0.0, 3.5],
        [3.0 + 1.5 * np.sqrt(3), 0.0, 3.5 + 2.0 * np.sqrt(3)],
        [3.0, 0.0, 2.0 + 2.0 * np.sqrt(3)],
    ]
```

imas_paraview/plugins/pf.py
```diff
@@ -0,0 +1,273 @@
...
+
+            logger.info(
+                f"Loaded {quantity_name!r} with {len(quantity.element)} element(s)."
+            )
+
+    def _create_outline(self, outline):
+        """Create vtkPolyData object from outline geometry"""
+        r, z = outline.r, outline.z
+        points = [(r_i, 0.0, z_i) for r_i, z_i in zip(r, z)]
+        return points_to_vtkpoly(points, is_closed=True)
```
maarten-ic: If I understand correctly, this will just show the outline, and no face is created?

maarten-ic (commented):
Couple of comments, let me know if you have any questions!

paulotex:
> > Something to consider; the following IDSs contain the same geometry types as used in `pf_active` and `pf_passive`:
> > 
> > * [ferritic](https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/ferritic.html#ferritic-object-axisymmetric)
> > * [ic_antennas](https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/ic_antennas.html#ic_antennas-antenna-module-strap-geometry)
> > * [iron_core](https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/iron_core.html#iron_core-segment-geometry)

> Would be nice if we can plot those indeed! `iron_core` is used for WEST and should be available in some of the WEST datasets on SDCC, perhaps you could test with that? As for the naming, `Axisymmmetric Geometry Reader` would not be immediately obvious to me when you'd need to choose this (e.g. wall 2D is also axisymmetric but cannot be read by this one). @paulotex do you have a preference or suggestion?

The common AoS is the "geometry"? If yes, Does "geometry reader" work, or is it too generic?

imas_paraview/plugins/pf.py
SBlokhuizen: I'll pick this up in a separate PR, as this is being done in many other files. I have updated the logging for just this file for now.

docs/source/pf_reader.rst
SBlokhuizen: There are no `pf_active` or `pf_passive` IDSs in the [machine description datasets](https://zenodo.org/records/15525525) on Zenodo. Is there other public data available?

imas_paraview/plugins/pf.py
SBlokhuizen: Exactly, this will just store the points and lines of the outline. If we want to store the faces as well, we could store them as [vtkPolygon](https://vtk.org/doc/nightly/html/classvtkPolygon.html)s in a [vtkPolyData](https://vtk.org/doc/nightly/html/classvtkPolyData.html) object.

SBlokhuizen:
> > > Something to consider; the following IDSs contain the same geometry types as used in `pf_active` and `pf_passive`:
> > > 
> > > * [ferritic](https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/ferritic.html#ferritic-object-axisymmetric)
> > > * [ic_antennas](https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/ic_antennas.html#ic_antennas-antenna-module-strap-geometry)
> > > * [iron_core](https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/iron_core.html#iron_core-segment-geometry)
> 
> > Would be nice if we can plot those indeed! `iron_core` is used for WEST and should be available in some of the WEST datasets on SDCC, perhaps you could test with that? As for the naming, `Axisymmmetric Geometry Reader` would not be immediately obvious to me when you'd need to choose this (e.g. wall 2D is also axisymmetric but cannot be read by this one). @paulotex do you have a preference or suggestion?
> 
> The common AoS is the "geometry"? If yes, Does "geometry reader" work, or is it too generic?

For the [ferritic](https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/ferritic.html#ferritic-object-axisymmetric) IDS it is just stored in an AoS called `axisymmetric`, but for the others they are indeed stored in the `geometry` AoS. 

I would argue that `Geometry Reader` is too vague. For example, there are different `geometry` structures available in other IDSs as well, for example:
- [camera_ir](https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/camera_ir.html#camera_ir-fibre_bundle-geometry)
- [ec_launcers](https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/ec_launchers.html#ec_launchers-mirror-geometry)

Thoughts @maarten-ic ?

imas_paraview/plugins/pf.py
maarten-ic: I think it makes more sense to show the face as well, maybe check with Paulo on this?

imas_paraview/plugins/pf.py
```diff
@@ -156,89 +171,95 @@ def _create_oblique(self, oblique):
...
 
-            # Center of the circle
-            cr = mr + curvature_sign * h * nr
-            cz = mz + curvature_sign * h * nz
+        t = np.linspace(0, 1, self.resolution, endpoint=False)
 
-            # Angles of endpoints relative to center
-            theta1 = np.arctan2(z1 - cz, r1 - cr)
-            theta2 = np.arctan2(z2 - cz, r2 - cr)
+        theta = theta1[:, None] + t[None, :] * (theta2 - theta1)[:, None]
```
maarten-ic: Why not
```python
theta = np.linspace(theta1, theta2, self.resolution, endpoint=False)
```

You may need to add a `.T` in case the axes are in the wrong order :)

imas_paraview/plugins/pf.py
```diff
@@ -156,89 +171,95 @@ def _create_oblique(self, oblique):
...
+    def _create_annulus(self, annulus):
         """Create vtkPolyData object from annulus geometry"""
         r0, z0 = annulus.r, annulus.z
         r_in, r_out = annulus.radius_inner, annulus.radius_outer
 
-        outer_ids = []
-        inner_ids = []
+        theta = np.linspace(0, 2.0 * np.pi, self.resolution, endpoint=False)
+        c = np.cos(theta)
+        s = np.sin(theta)
```
maarten-ic: Please don't use single-letter variables. It's actually more clear to put the `np.cos(theta)` in the expressions for inner/outer points:

```python
# Perhaps nice to introduce this helper function?
def _stack_r_z(r, z):
    return np.column_stack((r, np.zeros_like(r), z))

inner_points = _stack_r_z(r0 + r_in * np.cos(theta), z0 + r_in * sin(theta))
```

maarten-ic (commented):
Some small comments in the numpy logic, let me know if you have any questions

SBlokhuizen:
The last commits introduce the following changes to this reader, as discussed in the previous progress meeting:
- The reader now supports loading the axisymmetric geometries from the `ferritic`, `ic_antennas`, and `iron_core` IDSs.
- The reader was renamed from `PF Reader` to `Geometry Reader (Axisymmetric)`. The documentation, file names, class name, etc. are updated accordingly.
- The reader now exports the the axisymmetric geometries as polygons instead of outlines. If a user wants to visualize the outlines like in the previous solution, they can simply apply the [Feature Edges](https://www.kitware.com/new-in-paraview-feature-edges-representation/) ParaView filter to this reader.

maarten-ic (approved):
Nice work!

SBlokhuizen:
Hi @paulotex, this PR implements a reader for visualizing axisymmetric geometries. Could you try it out and let us know if you have any feedback on it? Thanks!

PR #42, opened 2026-01-16
Reader for coils_non_axisymmetric IDS

SBlokhuizen:
This introduces a new ParaView plugin: the `Non-Axisymmetric Coils Reader`.

This reader allows you to visualize the [coil conductor](https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/coils_non_axisymmetric.html#coils_non_axisymmetric-coil-conductor) geometries from the [coils_non_axisymmetric](https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/coils_non_axisymmetric.html#coils-non-axisymmetric) IDS. This has been tested on the following machine description IDSs:

| URI                                                        | Element Geometries                | Cross-section geometry |
|------------------------------------------------------------|-----------------------------------|------------------------|
| imas:hdf5?path=/work/imas/shared/imasdb/ITER_MD/3/111003/2 | `line segment`, `arc of a circle` | None                   |
| imas:hdf5?path=/work/imas/shared/imasdb/ITER_MD/3/111004/1 | `line segment`, `arc of a circle` | None                   |
| imas:hdf5?path=/work/imas/shared/imasdb/ITER_MD/3/115001/2 | `line segment`                    | `annular`              |
| imas:hdf5?path=/work/imas/shared/imasdb/ITER_MD/3/115002/2 | `line segment`                    | `annular`              |
| imas:hdf5?path=/work/imas/shared/imasdb/ITER_MD/3/115003/2 | `line segment`                    | `annular`              |

The [geometrical element type](https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/coils_non_axisymmetric.html#coils_non_axisymmetric-coil-conductor-elements-types) `full circle` has also been implemented, though I was not able to find any IDSs using this geometry type.


imas_paraview/plugins/coils_non_axisymmetric.py
```diff
@@ -0,0 +1,273 @@
...
+
+
+@smproxy.source(label="Non-Axisymmetric Coils Reader")
+@smhint.xml("""<ShowInMenu category="IMAS Tools" />""")
+class CoilsNonAxisymmetricReader(GGDVTKPluginBase):
+    def __init__(self):
+        super().__init__("vtkMultiBlockDataSet", SUPPORTED_IDS_NAMES)
+        self.selectable_map = {}
+        self.resolution = 10  # Resolution for arcs_of_circle and full_circle geometries
+        self.cs_resolution = 10  # Cross-sectional resolution
```
maarten-ic: Making these comments "docstrings" (it's not the same thing, but whatever) makes them documentable for Sphinx and language servers :)

```suggestion
        self.resolution = 10
        """Resolution for arcs_of_circle and full_circle geometries"""
        self.cs_resolution = 10
        """Cross-sectional resolution"""
```

imas_paraview/plugins/coils_non_axisymmetric.py
```diff
@@ -0,0 +1,273 @@
...
+                logger.warning(
+                    "Conductor element %d has unsupported element type %d, skipping",
+                    elem_idx,
+                    elem_type,
+                )
+                continue
+
+            vtk_conductor = points_to_vtkpoly(elem_points)
+
+            if len(conductor.cross_section) > 0:
```
maarten-ic: Log when there is no cross-section? I'd presume that some data would be missing in this case?

I like the fallback to just showing the centerline in this case :)

imas_paraview/plugins/coils_non_axisymmetric.py
```diff
@@ -0,0 +1,273 @@
...
+        """Create annular cross-section around a polyline using VTK tube filters.
+
+        Args:
+            poly_line: VTK polyline representing conductor element.
+            cross_section: Cross-section object containing inner radius and width.
+
+        Returns:
+            VTK polydata representing the annulus geometry.
+        """
+        outer_tube = vtkTubeFilter()
```
maarten-ic, line 353: 💯

imas_paraview/plugins/coils_non_axisymmetric.py
```diff
@@ -0,0 +1,273 @@
...
+            elem_idx: Index of the conductor element.
+
+        Returns:
+            VTK polydata with cross-section geometry.
+        """
+        cross_section_idx = 0 if len(conductor.cross_section) == 1 else elem_idx
+        cross_section = conductor.cross_section[cross_section_idx]
+
+        if cross_section.geometry_type.index == 5:  # Annulus
+            return self._add_annulus_cross_section(input_poly_line, cross_section)
```
maarten-ic: It should be easy to support a circle geometry as well, right?

imas_paraview/plugins/coils_non_axisymmetric.py
```diff
@@ -0,0 +1,273 @@
...
+            max_angle = 2 * np.pi
+            t = np.linspace(0, max_angle, self.resolution + 1, endpoint=True)[:, None]
+        else:
+            # Sweep circle arc from start to end point
+            p_end = self._pol_to_cart3d(elements.end_points, idx)
+            v_end = p_end - p_centre
+            max_angle = np.arctan2(np.dot(v_end, tangent), np.dot(v_end, v_start))
+            if max_angle < 0:
+                max_angle += 2 * np.pi
+            t = np.linspace(0, max_angle, self.resolution)[:, np.newaxis]
```
maarten-ic: Inconsistent use of `np.newaxis` vs. `None` in the first branch?

Can't you take the `np.linspace` out of the `if/else` and use the same logic in both cases (but with a different `max_angle`)?

imas_paraview/tests/test_coils_non_axisymmetric.py
```diff
@@ -0,0 +1,85 @@
...
+    elements.intermediate_points.phi = [np.pi / 2]
+    elements.intermediate_points.z = [5.0]
+
+    elements.end_points.r = [10.0]
+    elements.end_points.phi = [np.pi]
+    elements.end_points.z = [5.0]
+
+    elements.centres.r = [0.0]
+    elements.centres.phi = [0.0]
+    elements.centres.z = [5.0]
```
maarten-ic, line 64: Wondering: what happens when `|start - center| != |end - center|` (or `|intermediate - center|`)?

imas_paraview/plugins/coils_non_axisymmetric.py
```diff
@@ -0,0 +1,273 @@
...
+        self.selectable_map = {}
+        for i, coil in enumerate(self._ids.coil):
+            coil_name = coil.name
+            if not coil_name:
+                coil_name = f"coil {i}"
+                logger.warning(
+                    "Non-axisymmetric coil without name found. Using %r", coil_name
+                )
+            # Coil names are not unique in some machine description IDSs
+            coil_name = f"{coil_name} / {coil.identifier}"
```
maarten-ic: `coil.identifier` is only valid for DD 3.x, not for DD 4.x

maarten-ic (commented):
First review, got a couple of comments already, but I'm not fresh enough anymore so might have missed a couple of things.

I'll have another detailed look again in the next review round 😉 Can you first pick up the following? Let me know if you have any questions!

imas_paraview/tests/test_coils_non_axisymmetric.py
SBlokhuizen: I have updated the reader to skip and log cases where a circular element is ill-defined. For example, when the start, intermediate, and end points are not equidistant from the center point, or when the four points (start, intermediate, end, and center) are not in the same plane. Let me know what you think!

imas_paraview/plugins/coils_non_axisymmetric.py
```diff
@@ -151,14 +180,26 @@ def create_conductor_geometry(self, conductor):
...
                 )
 
             conductor_block.AddInputData(vtk_conductor)
+            has_input = True
+
+        if not has_input:
+            logger.warning(
+                "Conductor does not have any valid geometrical elements, skipping"
+            )
+            return None
```
maarten-ic, line 193: Isn't this log message duplicated with line +116?

imas_paraview/plugins/coils_non_axisymmetric.py
```diff
@@ -151,14 +180,26 @@ def create_conductor_geometry(self, conductor):
                 )
                 continue
 
+            if elem_points is None:
+                logger.warning(
+                    "element %d has an invalid geometrical element, skipping", elem_idx
```
maarten-ic: This is an element of a coil, right? But now I don't know which coil?

Perhaps use `imas.util.get_full_path(...)` to obtain the full path to the problematic element?

maarten-ic: Same applies to the other log statements. If you're repeating yourself often then consider writing a helper method

imas_paraview/plugins/coils_non_axisymmetric.py
```diff
@@ -209,17 +259,39 @@ def _create_circular_geometry(self, elements, idx, is_full_circle):
...
+                logger.warning(
+                    "Start and end point of element %d are not equidistant from "
+                    "the centre point",
+                    idx,
+                )
+                return None
+
+            if not self._are_points_coplanar(p_start, p_intermediate, p_end, p_centre):
+                logger.warning("Element %d points are not coplanar", idx)
+                return None
```
maarten-ic: Can't you check this in the same place where you do the check for the start/intermediate point?

I'd prefer to have the sanity checks all combined in one place, preferably near the start of this method :) 
```
# Sanity checks
if not isclose(radius, |p_intermediate - p_center|):
    log(...)
if not full_circle and not isclose(...):
    log(...)
if not full_circle and not coplanar(...):
    log(...)
```

imas_paraview/plugins/coils_non_axisymmetric.py
```diff
@@ -209,17 +259,39 @@ def _create_circular_geometry(self, elements, idx, is_full_circle):
...
-            t = np.linspace(0, max_angle, self.resolution)[:, np.newaxis]
+            resolution = self.resolution
+        t = np.linspace(0, max_angle, resolution)[:, None]
         return p_centre + radius * (np.cos(t) * v_start + np.sin(t) * tangent)
 
+    def _are_points_coplanar(self, p0, p1, p2, p3):
+        """Checks if 4 points are coplanar."""
+        n = np.cross(p1 - p0, p2 - p0)
+        if np.linalg.norm(n) == 0.0:
+            return False
```
maarten-ic: I'm not sure if this is correct?

When `p0 == p1`, then this cross product is zero and you return False, but three distinct points are always coplanar?

imas_paraview/plugins/coils_non_axisymmetric.py
```diff
@@ -234,40 +306,49 @@ def _add_cross_section(self, input_poly_line, conductor, elem_idx):
...
         outer_tube.SetNumberOfSides(self.cs_resolution)
         outer_tube.Update()
+        append.AddInputData(outer_tube.GetOutput())
 
-        inner_tube = vtkTubeFilter()
-        inner_tube.SetInputData(poly_line)
-        inner_tube.SetRadius(cross_section.radius_inner)
-        inner_tube.SetNumberOfSides(self.cs_resolution)
-        inner_tube.Update()
+        if cs_type == "annulus":
```
maarten-ic: Why don't you check `cross_section.geometry_type.index == 5` here, and make `_add_cross_section` simpler?

```python
        if cross_section.geometry_type.index in (2, 5):  # Circle or annulus
            return self._add_circular_cross_section(input_poly_line, cross_section)
```

maarten-ic (commented):
Couple more comments, let me know if anything's unclear!

imas_paraview/tests/test_coils_non_axisymmetric.py
maarten-ic: Nice, good to have this explicitly logged indeed!

imas_paraview/plugins/coils_non_axisymmetric.py
SBlokhuizen: This is true, this was implicitly checking whether any of these points were equal, which I don't think we should allow as this would create an invalid arc of circle element. I have now reformatted the validity checking of the points to make this more explicit.

imas_paraview/plugins/coils_non_axisymmetric.py
```diff
@@ -190,11 +222,20 @@ def _create_circular_geometry(self, elements, idx, is_full_circle):
...
         p_start = self._pol_to_cart3d(elements.start_points, idx)
         p_intermediate = self._pol_to_cart3d(elements.intermediate_points, idx)
         p_centre = self._pol_to_cart3d(elements.centres, idx)
+        p_end = (  # End point is only defined for arc of circle
+            None if is_full_circle else self._pol_to_cart3d(elements.end_points, idx)
+        )
+
+        if not self._are_circular_points_valid(
+            p_start, p_intermediate, p_end, p_centre, is_full_circle
+        ):
```
maarten-ic, line 237: `is_full_circle` <=> `p_end is None`, so you could leave out the `is_full_circle` argument. However, this is more explicit and probably more clear when someone has a look at this in the future, so fine to keep it this way.

maarten-ic (approved):
Thanks, LGTM!

SBlokhuizen:
Hi @paulotex, this PR implements a reader for visualizing geometries from the `coils_non_axisymmetric` IDS. Could you try it out and let us know if you have any feedback on it? Thanks!

PR #43, opened 2026-01-16
Add additional ruff lint rules

SBlokhuizen:
This enables the following linting rules in ruff:
```
# pycodestyle
"E",
# Pyflakes
"F",
# pyupgrade
"UP",
# flake8-bugbear
"B",
# flake8-simplify
"SIM",
# isort
"I",
# eradicate
"ERA",
# flake8-comprehensions
"C4",
# flake8-logging
"LOG",
# flake8-logging-format
"G",
# flake8-tidy-imports
"TID",
# flake8-use-pathlib
"PTH",
# Refurb
"FURB",
# Ruff specific rules
"RUF",
```

### DIFF
--- a/imas_paraview/convert.py
+++ b/imas_paraview/convert.py
@@ -50,17 +50,17 @@ class Converter:
             index_list: A list of time indices to convert. By default only the first
                 time step is converted.
         """
-        logger.info(f"Creating a output directory at {output_path}")
+        logger.info("Creating a output directory at %s", output_path)
         output_path.mkdir(parents=True, exist_ok=True)
         output_path = output_path / self.ids.metadata.name
         if any(index >= len(self.ids.time) for index in index_list):
             raise RuntimeError("A provided index is out of bounds.")
 
         for index in index_list:
-            logger.info(f"Converting time step {self.ids.time[index]}...")
+            logger.info("Converting time step %f...", self.ids.time[index])
             vtk_object = self.ggd_to_vtk(time_idx=index)
             if vtk_object is None:
-                logger.warning(f"Could not convert GGD at time index {index} to VTK.")
+                logger.warning("Could not convert GGD at time index %d to VTK.", index)
                 continue
             self._write_vtk_to_xml(vtk_object, Path(f"{output_path}_{index}"))
 
@@ -136,7 +136,7 @@ class Converter:
         Retrieve the GGD grid from a reference IDS path and replace the current grid.
         """
         path = self.grid_ggd.path
-        logger.info(f"Fetching the GGD grid from the reference: '{path}'")
+        logger.info("Fetching the GGD grid from the reference: %r", path)
         path = path.strip("#")
         ids_name, path = path.split("/", 1)
         if ids_name not in self.dbentry.factory.ids_names():
@@ -188,15 +188,18 @@ class Converter:
             else:
                 time_idx = indices[0]
             logger.info(
-                f"Converting timestep: t = {self.ids.time[time_idx]} at index = "
-                f"{time_idx}"
+                "Converting timestep: t = %f at index = %d",
+                self.ids.time[time_idx],
+                time_idx,
             )
             return time_idx
         else:
             time_idx = len(self.ids.time) // 2
             logger.info(
-                "No time or time index provided, so converting the middle time "
-                f"step: t = {self.ids.time[time_idx]} at index {time_idx}."
+                "No time or time index provided, so converting the middle time step: "
+                "t = %f at index %d.",
+                self.ids.time[time_idx],
+                time_idx,
             )
             return time_idx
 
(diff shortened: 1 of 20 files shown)


imas_paraview/convert.py
```diff
@@ -136,7 +136,7 @@ def _replace_grid_with_reference(self):
         Retrieve the GGD grid from a reference IDS path and replace the current grid.
         """
         path = self.grid_ggd.path
-        logger.info(f"Fetching the GGD grid from the reference: '{path}'")
+        logger.info("Fetching the GGD grid from the reference: %r", path)
```
maarten-ic: I expect this will log something like `<IDSString0D [...]`, which is not what you want
```suggestion
        logger.info("Fetching the GGD grid from the reference: '%s'", path)
```

pyproject.toml
```diff
@@ -66,3 +66,6 @@ local_scheme = "no-local-version"
 
 [tool.ruff]
 exclude = ["imas_paraview/_version.py"]
+
+[tool.ruff.lint]
+select = ["G004"]
```
maarten-ic: I haven't read in detail, but I think this will select ONLY this rule, and none other? Can you check please? https://docs.astral.sh/ruff/linter/#rule-selection

Also I'd prefer if we can just enable the whole category instead of a specific rule, can you check that as well?

maarten-ic (commented):
Thanks for the refactor! Can you check the `%r` formats and the pyproject.toml configs (see comments below)?

pyproject.toml
SBlokhuizen: You are right, this only enabled this specific rule! I have now updated the ruff rules. 

Instead of just extending ruff's default selection with the new logging rules, I propose we follow ruff's recommended guidelines:

> Prefer [lint.select](https://docs.astral.sh/ruff/settings/#lint_select) over [lint.extend-select](https://docs.astral.sh/ruff/settings/#lint_extend-select) to make your rule set explicit.

I have enabled a list of popular rules proposed in the [Ruff docs](https://docs.astral.sh/ruff/linter/#rule-selection), as well as the new logging rules (`"G"`). Besides tidying up some notation, this also picked out some some bugs that were introduced in the past, so I am a fan of the new rules :)!

Let me know what you think!

maarten-ic (commented):
Can you check with Yannick on which `ruff` rules to enable? E.g. with SimDB he enabled a couple more, see: https://github.com/iterorganization/SimDB/blob/develop/pyproject.toml#L132

SBlokhuizen:
> Can you check with Yannick on which `ruff` rules to enable? E.g. with SimDB he enabled a couple more, see: https://github.com/iterorganization/SimDB/blob/develop/pyproject.toml#L132

I discussed with Yannick and decided to enable the same set of rules as we do for SimDB.

imas_paraview/io/read_ps.py
```diff
@@ -283,13 +284,9 @@ def _add_aos_scalar_array_to_vtk_field_data(
...
                 except IndexError:
-                    logger.warning(
-                        f"           no index {i} for subset {subset_idx}..."
-                    )
+                    logger.warning("no index %d for subset %d...", i, subset_idx)
                 except AttributeError:
-                    logger.warning(
-                        f"           no index {i} for subset {subset_idx}..."
-                    )
+                    logger.warning("no index %d for subset %d...", i, subset_idx)
```
maarten-ic, line 287: If we do exactly the same thing in both cases, why not collapse this into one except?

```python
except (IndexError, AttributeError):
    ...
```
(perhaps you may even leave out the parentheses, but I'm not entirely sure about that)

imas_paraview/io/read_geom.py
```diff
@@ -176,9 +174,9 @@ def _fill_vtk_cell_array_from_gs(
                 .boundary
             )
 
-            pt_ids = list(
-                map(lambda val: val - 1, obj_nodes)
-            )  # offset by -1 as fortran indexing used in IMAS( 1,...n)
+            pt_ids = [
+                val - 1 for val in obj_nodes
+            ]  # offset by -1 as fortran indexing used in IMAS( 1,...n)
```
maarten-ic: While we're fixing this up, can you put the comment on its own line to avoid this ugly formatting? :)

```suggestion
            # offset by -1 as fortran indexing used in IMAS( 1,...n):
            pt_ids = [val - 1 for val in obj_nodes]
```

Perhaps also align the comment text across these similar statements (e.g. line +124, and +198)?

imas_paraview/plugins/vtkggdreader.py
```diff
@@ -20,16 +20,18 @@
...
-# described by the GGD guidelines. They are for now denoted as experimental IDS, as they
-# are currently not covered by the unit tests. If the DD for these stays like this, they
-# will need to be handled separately. GGD grid locations for each IDS:
-# equilibrium (grids_ggd/grid)
-# distribution_sources (source/ggd/grid)
-# distributions (distribution/ggd/grid)
-# tf (field_map/grid)
-# transport_solver_numerics (boundary_conditions_ggd/grid)
-# waves (coherent_wave/full_wave/grid)
+"""
```
maarten-ic: Could you explain why this must be a multiline string instead of a comment? 🤔

maarten-ic (approved):
Looking good! Just some nitpicking left 😅 but would be fine to merge regardless

imas_paraview/plugins/vtkggdreader.py
SBlokhuizen: [Eradicate](https://docs.astral.sh/ruff/rules/commented-out-code/) seems too think that the lines containing the IDSs are actually unused python code. Other alternatives are to add noqa to each line (or to the full file) or ensure they are invalid python code, e.g.:
```
-> equilibrium (grids_ggd/grid) 
-> distribution_sources (source/ggd/grid)
```
A multiline string seemed to be the least intrusive option to me.

PR #44, opened 2026-01-28
Add identifier number if AoS does not contain a label

SBlokhuizen:
If an AoS did not contain an identifier label (e.g. `identifier`, `name`, `label`), the same name would be generated for the different children of the AoS, and as a result earlier loaded data would be lost when loading it into ParaView. 

This adds an identifier number behind the AoS name to ensure uniqueness of the given name.

### DIFF
--- a/imas_paraview/ids_util.py
+++ b/imas_paraview/ids_util.py
@@ -103,21 +103,18 @@ def recursive_ggd_path_search(
         )
 
 
-def is_child_of_time_dependent_aos(node):
-    """Returns True if the provided IDS node is a child of a time-dependent AoS,
+def is_time_dependent_aos(node):
+    """Returns True if the provided IDS node is a time-dependent AoS,
     and False otherwise.
 
-    Examples of children of a time-dependent AoSs:
-    - equilibrium/time_slice[0]
-    - core_sources/source[0]/global_quantities[1]
-    - core_profiles/profiles_1d[2]
+    Examples of time-dependent AoSs:
+    - equilibrium/time_slice
+    - core_sources/source[0]/global_quantities
+    - core_profiles/profiles_1d
     """
-    parent = imas.util.get_parent(node)
     return (
-        hasattr(node.metadata, "coordinate1")
+        isinstance(node, IDSStructArray)
         and node.metadata.coordinate1.is_time_coordinate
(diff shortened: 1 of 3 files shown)


maarten-ic (commented):
~Looks good! Can you also add a test in `test_ids_util.py` for the new use case?~

Edit: sorry, initially reviewed the last commit only 😅

imas_paraview/ids_util.py
```diff
@@ -102,6 +103,24 @@ def recursive_ggd_path_search(
...
+    - core_sources/source[0]/global_quantities[1]
+    - core_profiles/profiles_1d[2]
+    """
+    parent = imas.util.get_parent(node)
+    return (
+        hasattr(node.metadata, "coordinate1")
+        and node.metadata.coordinate1.is_time_coordinate
+        and hasattr(parent.metadata, "coordinate1")
+        and parent.metadata.coordinate1.is_time_coordinate
+    )
```
maarten-ic:
```suggestion
def is_time_dependent_aos(node):
    """Returns True if the provided IDS node is a time-dependent AoS,
    and False otherwise.

    Examples of children of a time-dependent AoSs:
    - equilibrium/time_slice
    - core_sources/source[0]/global_quantities
    - core_profiles/profiles_1d
    """
    return isinstance(node, IDSStructArray) and node.metadata.coordinate1.is_time_coordinate
```

imas_paraview/ids_util.py
```diff
@@ -116,33 +135,34 @@ def create_name_recursive(node):
...
-
-        # Check if node has a label
-        elif hasattr(node, "label"):
-            name_appendix = str(node.label.value).strip()
-
-        # Add identifier/name/label in between brackets to the full name
-        if name_appendix != "":
+    parent = imas.util.get_parent(node)
+    # skip time slice quantity
+    if not is_child_of_time_dependent_aos(node):
```
maarten-ic:
```suggestion
    if not is_time_dependent_aos(parent):
```

maarten-ic (commented):
As just discussed, let's simplify the check for time-dependent AoS a bit. Please double-check the suggestions :)

maarten-ic (approved):
Thanks!

PR #45, opened 2026-02-04
Scalar Time Trace Reader

SBlokhuizen:
This PR introduces the `Scalar Time Trace Reader`. This reader can visualize time-dependent 0D scalar quantities in an IDS.


docs/source/scalar_time_trace.rst
```diff
@@ -0,0 +1,72 @@
...
+the 'Line Chart View' in ParaView. 
+See `Paraview's documentation <https://docs.paraview.org/en/latest/UsersGuide/displayingData.html#line-chart-view>`_ 
+to learn more about how to use this view.
+
+Within the 'Line Chart View' options, disable the ``Use Index For X Axis`` option, 
+and in the drop-down menu, select ``Time [s]`` to display time on the horizontal axis in the line chart.
+
+.. tip:: The Line Chart View can be opened side-by-side with other ParaView views. 
+   This means that this reader allows you to visualize how scalar quantities (e.g., the plasma current) 
+   evolve over time, while simultaneously displaying other time-dependent 3D geometries loaded from other readers.
```
maarten-ic:
```suggestion
   evolve over time, while simultaneously displaying time-dependent 3D geometries loaded from other readers.
```

docs/source/scalar_time_trace.rst
```diff
@@ -0,0 +1,72 @@
...
+
+This reader supports two ways of visualizing data:
+
+#. Partial Time Trace Up to Selected Time (Default)
+
+   When the ``Show Full Time Trace`` option is disabled in the plugin properties, 
+   the reader outputs only the time trace up to the currently selected time. 
+   This allows you to inspect the evolution of the selected quantities.
```
maarten-ic, line 57: Wouldn't it be nicer to show the images as part of this listing instead of really small in the table below?

imas_paraview/plugins/scalar_time_trace.py
```diff
@@ -0,0 +1,230 @@
...
+
+    def RequestData(self, request, inInfo, outInfo):
+        if self._dbentry is None or not self._ids_and_occurrence or self._ids is None:
+            return 1
+        if self._ids.ids_properties.homogeneous_time != IDS_TIME_MODE_HOMOGENEOUS:
+            logger.warning("Only IDSs with homogeneous time-mode are supported.")
+            return 1
+        if len(self._ids.time) == 0:
+            logger.warning("The IDS does not have a filled time array.")
+            return 1
```
maarten-ic, line 81: Should these be `logger.error` instead? Perhaps this is done elsewhere as well in case it'd be better to do in a separate PR :)

imas_paraview/plugins/scalar_time_trace.py
```diff
@@ -0,0 +1,230 @@
...
+
+        # Retrieve the selected time step
+        time = self._get_selected_time_step(outInfo)
+        if time is None:
+            logger.warning("Selected invalid time step")
```
maarten-ic, line 86:
```suggestion
            logger.error("Invalid time step selected")
```

imas_paraview/plugins/scalar_time_trace.py
```diff
@@ -0,0 +1,230 @@
...
+                inside a time-dependent AoS.
+            path_from_time_slice: Relative IDS path from the current time-dependent AoS
+                path to the current node.
+        """
+        metadata = node.metadata
+        # Skip time and GGD quantities
+        if metadata.name in ("time", "grid_ggd", "grids_ggd", "ggd", "description_ggd"):
+            return
+
+        parent = imas.util.get_parent(node)
```
maarten-ic: This feels more logical to do "one level higher" in your recursive algorithm. See comment below

imas_paraview/plugins/scalar_time_trace.py
```diff
@@ -0,0 +1,230 @@
...
+                    new_path = (
+                        f"{path_from_time_slice}/{subnode.metadata.name}"
+                        if path_from_time_slice
+                        else subnode.metadata.name
+                    )
+                self._recursively_find_time_dependent_quantities(
+                    subnode, time_slice=time_slice, path_from_time_slice=new_path
+                )
+                if is_time_dependent_aos(node):  # Only scan the first time slice
+                    break
```
maarten-ic: Let's simplify this? and get rid of the logic on lines 132-136?

Proposal in pseudo code:

```python
if node is IDSStructArray:
    if node.has_value and is_time_dependent_aos(node):
        assert time_slice is None
        self._recursively_find_time_dependent_quantities(node[0], time_slice=node, path_from_time_slice="")
    else:
        for i, subnode in enumerate(node):
            self._recursively_find_time_dependent_quantities(node[0], time_slice, f"{path_from_time_slice}[{i}]")
elif node is structure:
    new_path = ...
    self._recursively_find_time[...](...)
```

imas_paraview/plugins/scalar_time_trace.py
```diff
@@ -0,0 +1,230 @@
...
+        elif (
+            metadata.data_type in (IDSDataType.FLT, IDSDataType.INT)
+            and metadata.type == IDSType.DYNAMIC
+            and node.has_value
+            and (metadata.ndim == 0 or (metadata.ndim == 1 and time_slice is None))
```
maarten-ic: Should you also check that `time_slice is not None` when `metadata.ndim == 0`?

imas_paraview/plugins/scalar_time_trace.py
```diff
@@ -0,0 +1,230 @@
...
+        ):
+            name = f"{create_name_recursive(node)} [{node.metadata.units}]"
+            self._filled_quantities_map[name] = FilledQuantity(
+                node, time_slice, path_from_time_slice
+            )
+
+    def _load_time_dependent_data(self, output, selected_time):
+        """Load the selected time-dependent quantities up to the selected time.
```
maarten-ic: Update docstring?

imas_paraview/plugins/scalar_time_trace.py
```diff
@@ -0,0 +1,230 @@
...
+            itime = np.searchsorted(time_array, selected_time)
+            # Add duplicate row at selected time to prevent slanted time marker line
+            time_array = np.insert(time_array, itime, selected_time)
+            n_times = len(time_array) - 1
+        else:
+            time_indices = np.where(time_array <= selected_time)[0]
+            if len(time_indices) == 0:
+                logger.warning("No data available up to time %f", selected_time)
+                return
+            time_array = time_array[time_indices]
```
maarten-ic: This should work as well, right (and be slightly more efficient)?
```suggestion
            time_array = time_array[time_array <= selected_time]
            if len(time_array) == 0:
                logger.warning("No data available up to time %f", selected_time)
                return
```

imas_paraview/plugins/scalar_time_trace.py
```diff
@@ -0,0 +1,230 @@
...
+            quantity_array = self._get_quantity_array(node, n_times)
+
+            # Add duplicate row at selected time to prevent slanted time marker line
+            if self._show_full_time_trace:
+                quantity_array = np.insert(quantity_array, itime, quantity_array[itime])
+                ymin = min(ymin, np.nanmin(quantity_array))
+                ymax = max(ymax, np.nanmax(quantity_array))
+
+            self._add_column_to_table(output, quantity_array, quantity_name)
+            logger.info("Loaded '%s'", quantity_name)
```
maarten-ic: Maybe make this a debug message (or remove it?) to avoid overflowing the console with log messages?

imas_paraview/plugins/scalar_time_trace.py
```diff
@@ -0,0 +1,230 @@
...
+
+        # Add a marker line indicating the current time
+        if self._show_full_time_trace:
+            cursor = np.full(n_times + 1, np.nan)
+            cursor[itime] = ymin * 0.99
```
maarten-ic: What if `ymin` or `max` are negative?

If you want the marker to be slightly bigger than the range, it's probably nicer to do something like this?
```
cursor[itime] = ymin - (ymax - ymin) / 100
```

This will always scale nicely, whereas if `ymin = 9,999` and `ymax = 10,001` the current logic would add about 100 times the total data range as margin :)

maarten-ic (commented):
Added some comments, let me know if you have any questions!

imas_paraview/plugins/scalar_time_trace.py
SBlokhuizen: This happens in multiple other readers as well, so let's pick this up in a separate PR!

docs/source/scalar_time_trace.rst
```diff
@@ -55,18 +55,21 @@ This reader supports two ways of visualizing data:
    When the ``Show Full Time Trace`` option is disabled in the plugin properties, 
    the reader outputs only the time trace up to the currently selected time. 
    This allows you to inspect the evolution of the selected quantities.
+
+.. figure:: images/full_time_trace_disabled.png
+
+   Plasma current over time with ``Show Full Time Trace`` disabled (default), where
+   the selected time step is 70s.
+
 #. Full Time Trace with Highlighted Current Time
```
maarten-ic, line 64: Numbering resets here, perhaps you need to indent the figure to the same level as the text inside the listing?

maarten-ic (commented):
Thanks! Just a documentation markup comment now

maarten-ic (approved):

SBlokhuizen:
Hi @paulotex This PR implements the `Scalar Time Trace Reader` that we have discussed in the previous progress meeting. With this reader you can visualize time-dependent scalar quantities in an IDS. 

Besides being able to see the scalar quantity evolve over time, we have also implemented a second visualization mode based on the suggestion from @hrdoktorx during the progress meeting. In this mode, the reader will output a time trace containing all time steps in the IDS, along with a marker highlighting the current time step. In [the documentation](https://imas-paraview.readthedocs.io/en/latest/scalar_time_trace.html) you will find instructions on how to enable this along with example images.

Please have a try plotting some data (you should be able to use any IDS as input for this reader), and let us know if you have any feedback. Thanks!

PR #47, opened 2026-02-16
Add support for VTK to GGD conversion

SBlokhuizen:
This introduces the CLI command `vtk2ggd` which can convert VTK objects into an IDS containing GGD grids. Note that there are strict limitations on the type of VTK data the converter expects, and how it is structured. This has been written in the documentation. 

Besides testing the conversion with a generated dummy GGD unit test, I also validated the conversion using real IDS datasets. For each case, I first ran the `ggd2vtk` command, then converted the resulting VTK files back to GGD using `vtk2ggd`, and visually inspected the GGD grid and its subsets in ParaView:
 
- SOLPS: `imas:hdf5?path=/work/imas/shared/imasdb/ITER_SCENARIOS/3/123356/1`, with `edge_profiles` IDS
- JOREK: `imas:hdf5?user=public;pulse=112111;run=2;database=ITER_DISRUPTIONS;version=4`, with `wall` IDS
- MD: `iter_md_wall_116100_2001.nc`from [Zenodo](https://zenodo.org/records/17113713), with `wall` IDS. (Conversion of this dataset takes quite long (>1 minute), which is why it was not included as a unit test).

### DIFF
--- a/docs/source/cli.rst
+++ b/docs/source/cli.rst
@@ -144,14 +144,14 @@ VTK File Format Requirements
 
 The ``vtk2ggd`` converter has specific requirements for the VTK input files it can process:
 
-* The converter expects VTK Partitioned Dataset Collection (``.vtpc``) files in XML format
-* The converter can handle either:
+* The converter can convert the following VTK files in the XML format:
   
-  - A single ``.vtpc`` file containing a single time step conversion
+  - A single ``.vtu`` file containing a vtkUnstructuredGrid.
+  - A single ``.vtpc`` file containing a vtkPartitionedDataSetCollection with a single time step
   - A directory containing a file series of ``.vtpc`` files (e.g., ``edge_profiles_0.vtpc``, 
     ``edge_profiles_1.vtpc``, etc.), one for each time step.
 
-* Each ``vtkPartitionedDataSet`` should contain **exactly one partition** containing a ``vtkUnstructuredGrid``, otherwise the 
+* When providing ``.vtpc`` files, each ``vtkPartitionedDataSet`` should contain **exactly one partition** containing a ``vtkUnstructuredGrid``, otherwise the 
   partition is skipped. Each ``vtkUnstructuredGrid`` will be converted into a separate grid subset in the GGD.
 
 Limitations of VTK grid conversion
@@ -162,6 +162,6 @@ The converter can only output IDSs with:
 * a `linear <https://imas-data-dictionary.readthedocs.io/en/latest/generated/identifier/ggd_identifier.html>`_ GGD grid.
 * a single `space of a GGD grid <https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/edge_profiles.html#edge_profiles-grid_ggd-space>`_.
 * `standard space geometries <https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/edge_profiles.html#edge_profiles-grid_ggd-space-geometry_type>`_, not Fourier spaces.
-* X, Y, Z coordinates (`coordinate_identifier.x/y/z <https://imas-data-dictionary.readthedocs.io/en/latest/generated/identifier/coordinate_identifier.html>`_).
+* X, Y, Z coordinates (`coordinate_identifier.x/y/z <https://imas-data-dictionary.readthedocs.io/en/latest/generated/identifier/coordinate_identifier.html>`_), or R, Phi, Z coordinates (when the ``--cylindrical_coordinates`` flag is enabled).
 * the GGD grid structure itself, not any physical quantities (temperature, density, etc.) defined on the grids in the `ggd <https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/edge_profiles.html#edge_profiles-ggd>`_ AoS.
 * a homogeneous time mode.
--- a/docs/source/cli.rst
+++ b/docs/source/cli.rst
@@ -152,7 +152,7 @@ The ``vtk2ggd`` converter has specific requirements for the VTK input files it c
     ``edge_profiles_1.vtpc``, etc.), one for each time step.
 
 * Each ``vtkPartitionedDataSet`` should contain **exactly one partition** containing a ``vtkUnstructuredGrid``, otherwise the 
-  partition is skipped. Each ``vtkUnstructuredGrid`` will be converted into a separate grid subset in the GGD
+  partition is skipped. Each ``vtkUnstructuredGrid`` will be converted into a separate grid subset in the GGD.
 
 Limitations of VTK grid conversion
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/docs/source/cli.rst
+++ b/docs/source/cli.rst
@@ -107,3 +107,61 @@ To convert all time steps that fall between 2.2s and 6.6s:
 .. note:: If the specified time step is not found in the IDS, the time step before the
    specified time step will be used instead.
 
+
+Convert VTK to GGD grids using the CLI
+--------------------------------------
+
+IMAS-ParaView supports converting VTK files to GGD grids in an IDS using the ``vtk2ggd`` command-line tool.
+
+.. tip:: Detailed usage of the CLI tool can be found by running ``vtk2ggd --help``
+
+Usage
+^^^^^
+
+The ``vtk2ggd`` command requires two arguments: the path to VTK files and the output URI where the IDS will be stored.
+
+**Single file conversion:**
+
+.. code-block:: bash
+
+    vtk2ggd example.vtpc imas:hdf5?path=output_db --ids_name edge_profiles
+
+This converts a single ``.vtpc`` file to an IDS, at the given URI.
+
+**File series conversion:**
+
+.. code-block:: bash
+
+    vtk2ggd output_dir/ imas:hdf5?path=output_db
+
+This converts all ``.vtpc`` files in the specified directory to a single IDS with multiple time steps. The files
+should be named following the pattern ``{ids_name}_0.vtpc``, ``{ids_name}_1.vtpc``, etc. The IDS name will be 
+automatically inferred from the file names.
+
+
+VTK File Format Requirements
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The ``vtk2ggd`` converter has specific requirements for the VTK input files it can process:
+
+* The converter expects VTK Partitioned Dataset Collection (``.vtpc``) files in XML format
+* The converter can handle either:
+  
+  - A single ``.vtpc`` file containing a single time step conversion
+  - A directory containing a file series of ``.vtpc`` files (e.g., ``edge_profiles_0.vtpc``, 
+    ``edge_profiles_1.vtpc``, etc.), one for each time step.
+
+* Each ``vtkPartitionedDataSet`` should contain **exactly one partition** containing a ``vtkUnstructuredGrid``, otherwise the 
+  partition is skipped. Each ``vtkUnstructuredGrid`` will be converted into a separate grid subset in the GGD
+
+Limitations of VTK grid conversion
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The converter can only output IDSs with:
+
+* a `linear <https://imas-data-dictionary.readthedocs.io/en/latest/generated/identifier/ggd_identifier.html>`_ GGD grid.
+* a single `space of a GGD grid <https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/edge_profiles.html#edge_profiles-grid_ggd-space>`_.
+* `standard space geometries <https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/edge_profiles.html#edge_profiles-grid_ggd-space-geometry_type>`_, not Fourier spaces.
+* X, Y, Z coordinates (`coordinate_identifier.x/y/z <https://imas-data-dictionary.readthedocs.io/en/latest/generated/identifier/coordinate_identifier.html>`_).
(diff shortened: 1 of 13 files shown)


imas_paraview/tests/conftest.py
```diff
@@ -65,7 +65,13 @@ def pytest_sessionstart(session):
     with imas.DBEntry("netcdf_testdb.nc", "w", dd_version=DD_VERSION) as dbentry:
         dbentry.put(ids)
 
-    if not imas.backends.imas_core.imas_interface.has_imas:
+    try:
+        has_imas = imas.backends.imas_core.imas_interface.has_imas
+    except AttributeError:
+        # imas-python >= v2.2.0 always has IMAS-Core installed
+        has_imas = True
```
Yannicked: You could make a function out of this

imas_paraview/util.py
```diff
@@ -248,3 +250,41 @@ def points_to_vtkpoly(points, is_closed=False, is_filled=False):
...
+    reader.UpdateInformation()
+    out_info = reader.GetOutputInformation(0)
+
+    time_value = None
+    if out_info.Has(vtkStreamingDemandDrivenPipeline.TIME_STEPS()):
+        time_steps = out_info.Get(vtkStreamingDemandDrivenPipeline.TIME_STEPS())
+        if time_steps:
+            # NOTE: IMAS-ParaView exports only a single time step per
+            # vtkPartitionedDataSetCollection
+            time_value = time_steps[0]
```
Yannicked, line 302: Should you assert that there is only a single timestep?

Yannicked (approved):

docs/source/cli.rst
```diff
@@ -107,3 +107,61 @@ To convert all time steps that fall between 2.2s and 6.6s:
...
+
+* The converter expects VTK Partitioned Dataset Collection (``.vtpc``) files in XML format
+* The converter can handle either:
+  
+  - A single ``.vtpc`` file containing a single time step conversion
+  - A directory containing a file series of ``.vtpc`` files (e.g., ``edge_profiles_0.vtpc``, 
+    ``edge_profiles_1.vtpc``, etc.), one for each time step.
+
+* Each ``vtkPartitionedDataSet`` should contain **exactly one partition** containing a ``vtkUnstructuredGrid``, otherwise the 
+  partition is skipped. Each ``vtkUnstructuredGrid`` will be converted into a separate grid subset in the GGD
```
maarten-ic:
```suggestion
  partition is skipped. Each ``vtkUnstructuredGrid`` will be converted into a separate grid subset in the GGD.
```

imas_paraview/tests/conftest.py
maarten-ic: For tests we can assume that we're using the latest IMAS-Python, which comes for sure with IMAS-Core.

imas_paraview/cli.py
```diff
@@ -183,6 +184,79 @@ def convert_ggd_to_vtk(
...
+        vtpc_files = sorted(
+            path.glob("*.vtpc"),
+            key=lambda x: int(x.stem.split("_")[-1]) if "_" in x.stem else 0,
+        )
+        if not vtpc_files:
+            raise click.UsageError(f"No .vtpc files found in {path}")
+
+        # Infer ids_name from the first file (edge_profiles_0.vtpc -> edge_profiles)
+        if not ids_name:
+            ids_name = vtpc_files[0].stem.rsplit("_", 1)[0]
```
maarten-ic: Should you check that all vtpc files refer to the same IDS?

imas_paraview/cli.py
```diff
@@ -183,6 +184,79 @@ def convert_ggd_to_vtk(
...
+
+    with imas.DBEntry(uri, "x", dd_version=dd_version) as entry:
+        # Set version_put properties (version_put was added in DD 3.22)
+        if hasattr(ids.ids_properties, "version_put"):
+            ids.ids_properties.version_put.access_layer = (
+                entry._dbe_impl.access_layer_version()
+            )
+            version_put = ids.ids_properties.version_put
+            version_put.data_dictionary = entry.dd_version
+            version_put.access_layer_language = f"IMAS-Python {imas.__version__}"
```
maarten-ic: I don't understand why this would be needed? IMAS-Python already does this (and would overwrite any values you put here regardless)?

imas_paraview/vtk2ggd.py
```diff
@@ -0,0 +1,257 @@
...
+        """
+        # NOTE: We only support creating a single GGD space
+        grid_ggd.space.resize(1)
+        space = grid_ggd.space[0]
+        # NOTE: We only support primary standard spaces
+        space.identifier = identifiers.ggd_space_identifier.primary_standard
+        # NOTE: We only support non-fourier grids
+        space.geometry_type.index = 0
+
+        # NOTE: We only support X,Y,Z coordinates
```
maarten-ic: Maybe check with IO: we could support r,phi,z coordinates as well?

imas_paraview/vtk2ggd.py
```diff
@@ -0,0 +1,257 @@
...
+            space: GGD space structure to populate.
+            vtk_pdsc: vtkPartitionedDataSetCollection to read data from.
+        """
+        points, objects_per_dimension = self.extract_data_from_pdsc(vtk_pdsc)
+
+        # Fill nodes
+        if points is not None:
+            space.objects_per_dimension[0].object.resize(len(points))
+            for i, p in enumerate(points):
+                obj = space.objects_per_dimension[0].object[i]
```
maarten-ic: You could also do `for point, object in zip(points, space.[...].object):`

Please prefer more descriptive local variables

imas_paraview/vtk2ggd.py
```diff
@@ -0,0 +1,257 @@
...
+                obj.geometry.resize(3)
+                # GGD2VTK converts coordinates differently for wall than from other IDSs
+                if self.ids_name == "wall":
+                    obj.geometry[:] = [p[0], p[1], p[2]]
```
maarten-ic: Should you then also swap coordinate identifiers for wall?

imas_paraview/vtk2ggd.py
```diff
@@ -0,0 +1,257 @@
...
+            vtk_pdsc: vtkPartitionedDataSetCollection to read data from.
+        """
+        points, objects_per_dimension = self.extract_data_from_pdsc(vtk_pdsc)
+
+        # Fill nodes
+        if points is not None:
+            space.objects_per_dimension[0].object.resize(len(points))
+            for i, p in enumerate(points):
+                obj = space.objects_per_dimension[0].object[i]
+                obj.geometry.resize(3)
```
maarten-ic: I'd just assign a numpy array (or list) directly

```python
if self.ids_name != "wall":
    point = [point[0], point[2], point[1]]
obj.geometry = point
```

imas_paraview/vtk2ggd.py
```diff
@@ -0,0 +1,257 @@
...
+                cell_dim = 0
+                n_elem = ugrid.GetNumberOfPoints()
+            else:
+                cell_dim = ugrid.GetCell(0).GetCellDimension() if n_cells > 0 else 0
+                n_elem = n_cells
+            subset.dimension = cell_dim + 1
+
+            identifier_name = name.lower().replace("-", "_").replace(" ", "_")
+            if identifier_name in [m.name for m in identifiers.ggd_subset_identifier]:
+                subset.identifier = identifiers.ggd_subset_identifier[identifier_name]
```
maarten-ic: Wouldn't it be easier to assume that the identifier name exists, and handle an exception if not?
```
try:
    subset.identifier = identifiers.ggd_subset_identifier[identifier_name]
except XYZ:
```

imas_paraview/vtk2ggd.py
```diff
@@ -0,0 +1,257 @@
...
+            else:
+                subset.identifier.name = identifier_name
+                subset.identifier.index = 0
+                subset.identifier.description = "Unknown subset identifier"
+
+            subset.element.resize(n_elem)
+            for i in range(n_elem):
+                el = subset.element[i]
```
maarten-ic:
```suggestion
            for i, el in enumerate(subset.element):
```

maarten-ic (commented):
Some comments and suggestions, let me know if anything's unclear!

imas_paraview/vtk2ggd.py
SBlokhuizen: Good idea! Let's discuss with IO during the next meeting.

imas_paraview/cli.py
```diff
@@ -215,27 +215,37 @@ def convert_vtk_to_ggd(path, uri, ids_name, dd_version):
...
+            for vtpc_file in vtpc_files:
+                ids_names.add(vtpc_file.stem.rsplit("_", 1)[0])
+                if len(ids_names) > 1:
+                    raise click.UsageError(
+                        "Could not infer IDS name from file names, because there are "
+                        "multiple IDS names in the directory. Use '--ids_name' to set "
+                        "the required IDS name"
+                    )
+
+            ids_name = next(iter(ids_names))
```
maarten-ic: Tuple unpacking looks a bit cleaner here?

```python
(ids_name,) = ids_names
```

imas_paraview/io/read_geom.py
```diff
@@ -65,37 +67,60 @@ def fill_vtk_points(
...
+
+    supported = {X, Y, Z, R, PHI}
+    unsupported = set(coord_indices) - supported
+    if unsupported:
+        logger.error(
+            "Unsupported coordinate types in '%s', space[%d]: %s. They will be ignored",
+            ids_name,
+            space_idx,
+            unsupported,
+        )
```
maarten-ic, line 107: There was a bug with older versions of the Fortran GGD library which used different coordinate identifiers. You'll be able to observe this in older SOLPS-ITER datasets such as `imas:hdf5?path=/work/imas/shared/imasdb/ITER/3/123217/1`

Can you test with this URI?

We may need to add some logic here to try to detect this old GGD library (code/library structure) and handle that gracefully.

SBlokhuizen:
The last 5 commits introduce the following changes, based on the discussions we had yesterday:
- You can now pass `--cylindrical_coordinates` to the `vtk2ggd` command in order to store the output GGD grid in cylindrical coordinates instead of cartesian.
- The dummy grid that is used for unit tests was made asymmetric so we can detect when coordinates are flipped, e.g. `x` <--> `y`
- The `vtk2ggd` command now allows that you provide a single `.vtu` file as input.


> There was a bug with older versions of the Fortran GGD library which used different coordinate identifiers. You'll be able to observe this in older SOLPS-ITER datasets such as imas:hdf5?path=/work/imas/shared/imasdb/ITER/3/123217/1
> 
> Can you test with this URI?
> 
> We may need to add some logic here to try to detect this old GGD library (code/library structure) and handle that gracefully.

@maarten-ic The SOLPS grids I tested stored the grids in r-phi coordinates instead of r-z. As r-phi grids are in principle allowed, I added a custom check that tries to handle this case based on the `geometry_type.description`, which were the same for the SOLPS cases I tested. 

I have tested this on the 4 different types of SOLPS files that I could find with `scenario-summary` (as well as the one you mentioned).
- SOLPS-ITER: `imas:hdf5?path=/work/imas/shared/imasdb/ITER/3/123345/1`
- SOLPS4.3: `imas:hdf5?path=/work/imas/shared/imasdb/ITER/3/122403/3`
- SOLPS-ITER_WG: `imas:hdf5?path=/work/imas/shared/imasdb/ITER/4/123364/1` (this had correct coordinate identifiers already)
- SOLPS: `imas:hdf5?path=/work/imas/shared/imasdb/ITER/3/102491/3`

imas_paraview/io/read_geom.py
```diff
@@ -81,6 +81,24 @@ def fill_vtk_points(
...
     R, PHI = coord_id.r.value, coord_id.phi.value
 
+    # Old version of GGD Fortran library (<=1.12.0) did not set coordinate identifiers
+    # correctly, setting the points in r,phi-coordinates instead of r,z. For this case
+    # we overwrite the coordinate identifiers.
+    # This issue has been fixed in the following commit:
+    # https://github.com/iterorganization/GGD/commit/23af2f113e550fa6e8d05c982ddae53bf29c1cf1 # noqa: E501
+    if grid_ggd.space[
+        space_idx
+    ].geometry_type.description == "Poloidal plane cross-section" and coord_indices == [
```
maarten-ic: Maybe check if `"poloidal" in geometry.description.lower()`? That seems more general.

Perhaps @hrdoktorx can give his opinion on this as well?

hrdoktorx: I do not know what `geometry.description.lower` refers to. So I cannot comment on that aspect.
What are you try to detect?
Shouldn't you scan the coordinates used for that space?

maarten-ic: Sorry @hrdoktorx , I should have provided more context indeed:

_The problem:_
In older SOLPS-ITER datasets (before GGD library version 1.13) the grid_ggd.space.coordinates_type were filled with "4, 5" for `r, z`, whereas the[ Data Dictionary says that 4=r and 5=phi ](https://imas-data-dictionary.readthedocs.io/en/latest/generated/identifier/coordinate_identifier.html#identifier-utilities-coordinate_identifier.xml).

_Current solution and suggestion:_
This code tries to detect this situation and interpret the data as intended (i.e. `r, z`, not `r, phi`). Currently the code checks if the `geometry_type.description` is exactly equal to `"Poloidal plane cross-section"`. My suggestion was to check if the word `"poloidal"` is part of `geometry_type.description` converted to lower-case characters.

_Question:_
Do you think this is a good approach? Or do you prefer to detect (and work around) this situation in another way?

hrdoktorx: More simply, you can look at `space[].geometry_type.name` and check if that is "Poloidal".

SBlokhuizen: I have now updated it to check against  `space[].geometry_type.name`, as  @hrdoktorx suggested.

maarten-ic (approved):

PR #48, opened 2026-02-18
Fix error when loading empty GGD grid

SBlokhuizen:
Previously the GGD reader would raise an error when trying to load an empty GGD grid. This has now been handled more gracefully.

### DIFF
--- a/imas_paraview/util.py
+++ b/imas_paraview/util.py
@@ -97,7 +97,6 @@ def get_grid_ggd(ids, time=0.0, parent_idx=0):
             continue
         elif isinstance(node, IDSStructArray):
             if len(node) == 0:
-                logger.error("'%s' array of structures is empty.")
                 return None
 
             # Time dependent array of structure
--- a/imas_paraview/util.py
+++ b/imas_paraview/util.py
@@ -69,7 +69,7 @@ def get_ggd_path(ids_metadata) -> Optional[str]:
     return None
 
 
-def get_grid_ggd(ids, time=0, parent_idx=0):
+def get_grid_ggd(ids, time=0.0, parent_idx=0):
     """Finds and returns the grid_ggd within IDS at the time index ggd_idx. If the
     grid_ggd at ggd_idx time index does not exist, it tries to return the first
     grid_ggd. If this does not exist, it returns None.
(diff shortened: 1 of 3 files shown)


imas_paraview/util.py
```diff
@@ -103,13 +103,18 @@ def get_grid_ggd(ids, time=0, parent_idx=0):
...
                 node = node[0]
                 logger.warning(
                     "The GGD grid was not found at time index %d, so first "
                     "grid was loaded instead.",
                     ggd_idx,
                 )
         else:
+            if len(node) == 0:
+                return None
             node = node[parent_idx]
```
Yannicked: Might be a bit cleaner if you structure the loop like this:
```suggestion
    for path in grid_path.split("/"):
        node = node[path]
        if node.metadata.ndim == 0:
            continue  # Current node is a structure
        if len(node) == 0:
            return None

        if node.metadata.coordinate1.is_time_coordinate:
            # Time dependent array of structure
            # Let IMAS-Python handle the time mode (homogeneous/heterogeneous):
            time_array = node.coordinates[0]
            # Load closest previous time index
            ggd_idx = np.searchsorted(time_array, time, side="right") - 1

            if 0 <= ggd_idx < len(node):
                node = node[ggd_idx]
            else:
                node = node[0]
                logger.warning(
                    "The GGD grid was not found at time index %d, so first "
                    "grid was loaded instead.",
                    ggd_idx,
                )
        else:
            node = node[parent_idx]
```

SBlokhuizen: I have refactored it a bit to make it cleaner and handled explicitly handled some edge cases. Let me know what you think!

imas_paraview/util.py
```diff
@@ -86,31 +87,44 @@ def get_grid_ggd(ids, time=0, parent_idx=0):
...
-            pass  # Current node is a structure
-        elif node.metadata.coordinate1.is_time_coordinate:
-            # Time dependent array of structure
-            # Let IMAS-Python handle the time mode (homogeneous/heterogeneous):
-            time_array = node.coordinates[0]
-            # Load closest previous time index
-            ggd_idx = np.searchsorted(time_array, time, side="right") - 1
+        if isinstance(node, IDSStructure):
+            continue
+        elif isinstance(node, IDSStructArray):
```
maarten-ic: Node can be either an IDSStructure or an IDSStructArray, so you shouldn't need to check `isinstance(node, IDSStructArray)` anymore, right?

maarten-ic (approved):

PR #49, opened 2026-02-19
Add example images for readers and gallery

SBlokhuizen:
This adds images for example visualization for the following readers:
- Geometry Reader (Axisymmetric)
- Non-Axisymmetric Coils Reader

This also introduces a 'Gallery' section to the docs, in which visualizations can be added to showcase the capabilities of IMAS-ParaView


docs/source/images/non_axisymmetric.png
Yannicked, line 1: Why is this slightly tilted?

docs/source/index.rst
```diff
@@ -26,6 +26,7 @@ Manual
 
    self
    installing
+   gallery
```
maarten-ic, line 34: Maybe the gallery should be its own category instead of being a subitem of "Getting started"?

Image

maarten-ic (approved):

PR #52, opened 2026-03-10
Pellets Reader

SBlokhuizen:
Introduces a new reader which can visualize shattered pellet fragments from the [spi](https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/spi.html#spi-injector-fragment) IDS.

This was tested on the following URI:
`imas:hdf5?user=public;pulse=113113;run=2;database=ITER_DISRUPTIONS;version=4`

### DIFF
--- a/imas_paraview/plugins/pellets.py
+++ b/imas_paraview/plugins/pellets.py
@@ -1,4 +1,4 @@
-"""ParaView plugin to visualize SPI fragment positions"""
+"""ParaView plugin to visualize pellets"""
 
 import logging
 from dataclasses import dataclass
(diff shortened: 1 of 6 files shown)


imas_paraview/plugins/pellets.py
```diff
@@ -0,0 +1,315 @@
...
+    def _populate_fragments(self, injector, injector_name):
+        """Populate the selectable_map with shattered pellet fragments of a shattered
+        pellet injector.
+
+        Args:
+            injector: IDSStructure of a shattered pellet injector.
+            injector_name: Name of the injector.
+        """
+        if len(injector.fragment) > 0:
+            fragment = injector.fragment[0]
```
Yannicked: Can you explain why you're only looking at the first fragment?

Yannicked (approved):

PR #59, opened 2026-04-07
Add Camera Reader

SBlokhuizen:
This adds a new reader, the `Camera Reader`, to visualize the camera sightlines of `camera_ir` and `camera_visible`. This includes the option to match the ParaView camera with a selected camera, and get a point of view visualization of what the camera is seeing.


paulotex:
tested and working for me.

One request though:
When selecting the camera to snap the view, is it possible to mark the frustum in any way? Perhaps change the line color, or increase the thickness? For example, change the color to one of these:

image

imas_paraview/plugins/camera.py
```diff
@@ -0,0 +1,296 @@
...
+        # NOTE: This requires DD version >= 4.1.0
+        for ch_idx, channel in enumerate(self._ids.channel):
+            channel_name = str(channel.name) or f"channel {ch_idx}"
+
+            for cam_idx, camera in enumerate(channel.camera):
+                camera_name = str(camera.name) or f"camera {cam_idx}"
+                geometry = CameraGeometry(
+                    origin=np.array(
+                        [camera.pinhole.x, camera.pinhole.y, camera.pinhole.z]
+                    ),
```
Yannicked: It might be neat to define a helper function that puts the xyz coordinates of an object into a numpy array, such as: 
```python
def xyz(obj):
    return np.array([obj.x, obj.y, obj.z])
```

SBlokhuizen: This happens all over the code base, so I would suggest if we want to implement this to do it in a separate PR so we can tackle it in other places as well.

SBlokhuizen:
> tested and working for me.
> 
> One request though: When selecting the camera to snap the view, is it possible to mark the frustum in any way? Perhaps change the line color, or increase the thickness? For example, change the color to one of these:
> image


Hi @paulotex, I have tried to implement this but I cannot find a nice way to do it. The issue is that the readers just provide the VTK data for ParaView to display, how ParaView renders this (e.g. color / line width) we don't have much a control over from the plugin side. 

A workaround I have found is to attach an integer (0 or 1) to the camera data to flag the active selection. You would need to select this ``ActiveCamera`` array in the coloring properties though. What do you think of this solution?

https://github.com/user-attachments/assets/4c87a364-dfe6-46f2-9cc5-23364d1df0a6

Yannicked (approved):

docs/source/camera_reader.rst
```diff
@@ -0,0 +1,87 @@
...
+
+1. Load and apply the Camera Reader with at least one camera selected.
+2. In the **Snap View to Camera** section of the Properties panel, choose the desired
+   camera from the **Select Camera** drop-down menu.
+3. Press the **Snap View to Camera** button.
+
+The ParaView camera will immediately move to the position and orientation of the
+selected IDS camera.
+
+.. warning::
```
maarten-ic: Warning sounds too severe, perhaps use hint, note or tip? https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#admonitions-messages-and-warnings

imas_paraview/plugins/camera.py
```diff
@@ -0,0 +1,296 @@
...
+    propertygroup,
+    stringlistdomain,
+    stringvector,
+)
+from imas_paraview.plugins.base_class import GGDVTKPluginBase
+from imas_paraview.util import ensure_unique_name, pol_to_cart
+
+logger = logging.getLogger("imas_paraview")
+
+# TODO: add support for camera_x_rays IDS
```
maarten-ic, line 29: What's blocking this?

imas_paraview/plugins/camera.py
```diff
@@ -0,0 +1,296 @@
...
+        self.selectable_map: dict[str, CameraGeometry] = {}
+        self._snap_camera_name = ""
+        self._snap_requested = False
+
+    @stringvector(
+        name="SnapCameraList", information_only=1, si_class="vtkSIDataArrayProperty"
+    )
+    def P96_GetSnapCameraList(self):
+        """Return the list of loaded cameras for the snap dropdown."""
+        from vtkmodules.vtkCommonCore import vtkStringArray
```
maarten-ic: Why not make this a global import?

imas_paraview/plugins/camera.py
```diff
@@ -0,0 +1,296 @@
...
+    @stringlistdomain("SnapCameraList", name="snap_camera_list")
+    def P97_SetSnapCameraName(self, value):
+        """Select which loaded camera to snap the ParaView view to."""
+        self._snap_camera_name = str(value).strip()
+
+    @command_button_property("SnapToCamera", "Snap View to Camera", "P98_SnapToCamera")
+    def P98_SnapToCamera(self):
+        """Snap the ParaView camera to the selected camera."""
+        self._snap_requested = True
+        self.Modified()
```
maarten-ic: Could you add a comment why can't you directly call self._snap_view_to_gemoetry here?

imas_paraview/plugins/camera.py
```diff
@@ -0,0 +1,296 @@
...
+        if self._ids.metadata.name == "camera_ir":
+            self._extract_camera_ir()
+        else:  # camera_visible
+            self._extract_camera_visible()
+
+        self._selectable = list(self.selectable_map.keys())
+
+    def _extract_camera_ir(self):
+        """Extract camera geometries from camera_ir IDS."""
+        # NOTE: This requires DD version >= 4.1.0
```
maarten-ic: This isn't mentioned anywhere in the docs? What happens when users provide a camera_ir in older DD versions?

imas_paraview/plugins/camera.py
```diff
@@ -0,0 +1,296 @@
...
+        for ch_idx, channel in enumerate(self._ids.channel):
+            channel_name = str(channel.name) or f"channel {ch_idx}"
+
+            for cam_idx, camera in enumerate(channel.camera):
+                camera_name = str(camera.name) or f"camera {cam_idx}"
+                geometry = CameraGeometry(
```
maarten-ic, line 158: Should you verify that all required data fields are filled before trying to display the cameras?

imas_paraview/plugins/camera.py
```diff
@@ -0,0 +1,296 @@
...
+                        ]
+                    ),
+                )
+
+                geom_name = ensure_unique_name(
+                    f"{channel_name} / {camera_name}", list(self.selectable_map.keys())
+                )
+                self.selectable_map[geom_name] = geometry
+
+    def _extract_camera_visible(self):
```
maarten-ic, line 169: Great how two different IDSs have different ways to represent the camera geometries... 🤨

imas_paraview/plugins/camera.py
```diff
@@ -0,0 +1,296 @@
...
+            alpha = channel.viewing_angle_alpha_bounds
+            beta = channel.viewing_angle_beta_bounds
+
+            origin = np.array([x, y, centre.z])
+            forward = np.array(
+                [ap.x3_unit_vector.x, ap.x3_unit_vector.y, ap.x3_unit_vector.z]
+            )
+            up = np.array(
+                [ap.x2_unit_vector.x, ap.x2_unit_vector.y, ap.x2_unit_vector.z]
+            )
```
maarten-ic, line 206: Any reason that over here these are first declared in a local variable whereas in the other method you calculate these while constructing the CameraGeometry?

docs/source/plugins/camera_reader.rst
```diff
@@ -0,0 +1,87 @@
...
+
+The **Snap View to Camera** widget lets you align the active ParaView RenderView with
+any loaded camera so that the viewport matches the camera's optical position and
+field of view.
+
+To snap the view:
+
+1. Load and apply the Camera Reader with at least one camera selected.
+2. In the **Snap View to Camera** section of the Properties panel, choose the desired
+   camera from the **Select Camera** drop-down menu.
```
maarten-ic, line 57: I haven't tested this, but is a change in the dropdown communicated by ParaView to the plugin without clicking the "Apply" button of the properties panel?

imas_paraview/plugins/camera.py
```diff
@@ -0,0 +1,296 @@
...
+        for block_id, name in enumerate(self._selected):
+            geometry = self.selectable_map[name]
+            vtk_geom = self._build_camera_polydata(geometry)
+            if vtk_geom is not None:
+                output.SetBlock(block_id, vtk_geom)
+                meta = output.GetMetaData(block_id)
+                meta.Set(vtkCompositeDataSet.NAME(), name)
+                logger.info("Loaded frustum for '%s'.", name)
+
+    def _build_camera_polydata(self, geometry: CameraGeometry):
```
maarten-ic, line 243: Would perhaps be more logical as an instance method of the CameraGeometry class?

Though not very important so fine if you keep it as is.

imas_paraview/plugins/camera.py
```diff
@@ -0,0 +1,296 @@
...
+
+        c0 = base_center - half_h * right - half_v * up_ortho
+        c1 = base_center + half_h * right - half_v * up_ortho
+        c2 = base_center + half_h * right + half_v * up_ortho
+        c3 = base_center - half_h * right + half_v * up_ortho
+
+        points = np.array([geometry.origin, c0, c1, c2, c3])
+
+        vtk_pts = vtkPoints()
+        vtk_pts.SetData(numpy_to_vtk(points))
```
maarten-ic, line 274: Do you need to create the vtkPoints object, or does `polydata.SetPoints` accept the `numph_to_vtk(points)` output directly?

imas_paraview/plugins/camera.py
```diff
@@ -0,0 +1,296 @@
...
+            [2, 0, 2],  # origin to c1
+            [2, 0, 3],  # origin to c2
+            [2, 0, 4],  # origin to c3
+            [2, 1, 2],  # c0  to c1
+            [2, 2, 3],  # c1  to c2
+            [2, 3, 4],  # c2  to c3
+            [2, 4, 1],  # c3  to c0
+        ]
+
+        flat = np.array([v for edge in edges for v in edge], dtype=np.int64)
```
maarten-ic: Isn't this the same as:

```suggestion
        flat = np.array(edges, dtype=np.int64).flatten()
```

imas_paraview/plugins/camera.py
```diff
@@ -0,0 +1,296 @@
...
+    def _snap_view_to_geometry(self, geometry: CameraGeometry):
+        """Snap the currently active RenderView to the selected camera, setting
+        ParaView's camera to match with the selected camera's position, focal point,
+        and field of view.
+
+        Args:
+            geometry: The CameraGeometry to snap to.
+        """
+
+        from paraview.simple import GetActiveView
```
maarten-ic, line 305: Very nice that you can do this from a plugin!

Have you tested if this works when running ParaView in client <-> server mode? Cause in that case the plugin is running on the server side, while the viewpoint needs to be updated client-side.

imas_paraview/util.py
```diff
@@ -464,3 +464,16 @@ def create_vtk_arrows(positions, directions, scaling_factor=1.0):
     glyph.Update()
 
     return glyph.GetOutput()
+
+
+def ensure_unique_name(name, existing_names):
```
maarten-ic, line 469: Perhaps not for this PR, but can you unify the way that names and their uniqueness is handled for all (non-GGD?) plugins?

This is now handled slightly different for different readers:
- Most readers assume unique names (or do `"<object name> <aos_index>"` when the name is empty)
- Axisymmetric coil reader adds the `coil.identifier` to the name
- Axisymmetric geometry and camera reader now add `#1`, etc. for duplicate names

Maybe I'm missing more cases, `git grep 'name = '` could give you a starting point for looking into this.

maarten-ic (commented):
Couple of comments. Please have a look and let me know if you'd like to discuss any.
GH also indicates there's a merge conflict, can you have a look at that as well?

Thanks!

imas_paraview/plugins/camera.py
SBlokhuizen: We did not receive any data from IO yet for the `camera_x_rays` IDS, and since the data structure for these cameras is again different from the other two IDSs, I suggest we leave this until we have actual data.

imas_paraview/plugins/camera.py
SBlokhuizen: This will indeed not work `camera_ir` IDS older than 4.1.0, I actually could not find the required camera quantities in the older versions (e.g. [3.42.2-migration](https://imas-data-dictionary.readthedocs.io/en/3.42.2-migration/generated/ids/camera_ir.html)). I have not found any existing `camera_ir` IDSs in the MD database, so this was only tested with the IDSs which were generated using the Excel sheet provided by Paulo, which were version 4.1.0. 

Shall I just log an error if the DD version is too old, and add a note about this in the documentation?

imas_paraview/util.py
SBlokhuizen: Good idea, let's pick this up in a separate PR!

imas_paraview/plugins/camera.py
SBlokhuizen: SetPoints requires you to give a vtkPoints object, if you try to directly pass the vtk array it will complain:
`TypeError: SetPoints argument 1: method requires a vtkPoints, a vtkTypeFloat64Array was provided.
`

docs/source/plugins/camera_reader.rst
SBlokhuizen: No you will still need to press the apply button, I have updated the steps here to include that.

imas_paraview/plugins/camera.py
SBlokhuizen: Indeed, this will not work when you are running ParaView in client-server mode. I will add a note in the documentation and logging to handle this.

imas_paraview/plugins/camera.py
SBlokhuizen: No specific reason, no. I have updated it so they both declare the arrays as local variables first. I think that looks the most legible.

imas_paraview/plugins/camera.py
maarten-ic: Looks like these quantities should be present in 4.0.0 as well (they're not mentioned in the [upgrade guide](https://imas-data-dictionary.readthedocs.io/en/4.1.0/generated/changelog/ids_changes/4.0.0.html#camera-ir))

Agreed to log a nice error when the DD version is too old (otherwise you'll run into an AttributeError when trying to access nonexisting variables), and good to mention in the table of supported IDSs in the docs.

imas_paraview/plugins/camera.py
maarten-ic: Clear, thanks for checking!

imas_paraview/plugins/camera.py
SBlokhuizen: I think they are mentioned there under the `channel` AoS. I also cannot find the channel AoS when creating a new camera_ir IDS of version 4.0.0. Perhaps I am missing something? I have updated the docs and logged an error if DD version is below 4.1.0 for now. 

```
╭──────────────────────────── IDS: camera_ir (DD version 4.0.0) ────────────────────────────╮
│ Infrared camera for monitoring of Plasma Facing Components                                │
│ ╭───────────────────────────────────── Attributes ──────────────────────────────────────╮ │
│ │ has_value = False                                                                     │ │
│ │  metadata = <IDSMetadata for 'camera_ir'>                                             │ │
│ ╰───────────────────────────────────────────────────────────────────────────────────────╯ │
│ ╭───────────────────────────────────── Child nodes ─────────────────────────────────────╮ │
│ │         calibration = <IDSStructure (IDS:camera_ir, calibration)>                     │ │
│ │                code = <IDSStructure (IDS:camera_ir, code)>                            │ │
│ │        fibre_bundle = <IDSStructure (IDS:camera_ir, fibre_bundle)>                    │ │
│ │               frame = <IDSStructArray (IDS:camera_ir, frame with 0 items)>            │ │
│ │      frame_analysis = <IDSStructArray (IDS:camera_ir, frame_analysis with 0 items)>   │ │
│ │      ids_properties = <IDSStructure (IDS:camera_ir, ids_properties)>                  │ │
│ │             latency = <IDSFloat0D (IDS:camera_ir, latency, empty FLT_0D)>             │ │
│ │ latency_error_lower = <IDSFloat0D (IDS:camera_ir, latency_error_lower, empty FLT_0D)> │ │
│ │ latency_error_upper = <IDSFloat0D (IDS:camera_ir, latency_error_upper, empty FLT_0D)> │ │
│ │            midplane = <IDSStructure (IDS:camera_ir, midplane)>                        │ │
│ │                name = <IDSString0D (IDS:camera_ir, name, empty STR_0D)>               │ │
│ │     optical_element = <IDSStructArray (IDS:camera_ir, optical_element with 0 items)>  │ │
│ │                time = <IDSNumericArray (IDS:camera_ir, time, empty FLT_1D)>           │ │
│ ╰───────────────────────────────────────────────────────────────────────────────────────╯ │
╰───────────────────────────────────────────────────────────────────────────────────────────╯
```

maarten-ic: Yes you're right, my bad. Thanks for checking!

imas_paraview/plugins/camera.py
```diff
@@ -0,0 +1,363 @@
...
+        """Helper to check if a lazy-loaded cartesian IMAS vector quantity has data."""
+        try:
+            return vec.x.has_value and vec.y.has_value and vec.z.has_value
+        except AttributeError:
+            return False
+
+    def _extract_camera_ir(self):
+        """Extract camera geometries from camera_ir IDS."""
+        imas_version = self._ids.ids_properties.version_put.data_dictionary
+        if imas_version and Version(imas_version) < Version("4.1.0"):
```
paulotex, line 114: This check seems to be failing for me:

```
  File "/home/ITER/tribolp/Source/IMAS-ParaView/imas_paraview/plugins/camera.py", line 121, in _extract_camera_ir
    if imas_version and Version(imas_version) < Version("4.1.0"):
                        ^^^^^^^^^^^^^^^^^^^^^
  File "/work/imas/opt/EasyBuild/software/Python-bundle-PyPI/2023.10-GCCcore-13.2.0/lib/python3.11/site-packages/packaging/version.py", line 198, in __init__
    match = self._regex.search(version)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: expected string or bytes-like object, got 'IDSString0D'
```

paulotex:
> A workaround I have found is to attach an integer (0 or 1) to the camera data to flag the active selection. You would need to select this `ActiveCamera` array in the coloring properties though. What do you think of this solution?

Thanks, this would work. However, it now duplicates some of the functionality of the Multiblock Inspector. So I think we can leave this feature for now and we can point the user to the multiblock inspector, if they need to highlight a camera.

docs/source/plugins/camera_reader.rst
```diff
@@ -21,7 +21,7 @@ Currently, the following IDS and structures are supported in the Camera Reader:
    * - IDS
      - Structure
    * - ``camera_ir``
-     - `Camera channels <https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/camera_ir.html#camera_ir-channel>`__
+     - `Camera channels <https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/camera_ir.html#camera_ir-channel>`__ ``(DD version >=4.1.0)``
```
maarten-ic: Image

Doesn't look very pretty. Perhaps just write in regular text ("available in Data Dictionary 4.1.0 and later")?

imas_paraview/plugins/camera.py
```diff
@@ -101,36 +107,69 @@ def setup_ids(self):
 
         self._selectable = list(self.selectable_map.keys())
 
+    def _cart_vector_has_value(self, vec):
+        """Helper to check if a lazy-loaded cartesian IMAS vector quantity has data."""
+        try:
+            return vec.x.has_value and vec.y.has_value and vec.z.has_value
+        except AttributeError:
+            return False
```
maarten-ic: Might be better to move this function to the util  or imas-util module?

maarten-ic (commented):
Two minor comments. Can you have a look at the Merge conflicts as well?

Thanks!

maarten-ic (approved):

PR #61, opened 2026-04-10
Gallery improvements

SBlokhuizen:
This improves the workflow for adding new examples to the gallery. 

You can now add additional examples by creating a new directory in the  `gallery/examples` directory. This should contain an image, a YAML file with details about the example, and optionally a ParaView state file. The gallery page will then automatically be generated based on the available examples. 

Additionally, 2 new examples, including state files, which were created by Sven have been added to the gallery.

PR #62, opened 2026-04-10
Release 2.3.1

SBlokhuizen:
# Release notes IMAS-ParaView 2.3.1

## Changes since version 2.3.1

- Fix naming issue in the Profiles1DMapper (#57)
- Remove explicit dependency upon VTK (#56)
- Update SOLPS training URI to DDv4 (#55)
- Refactor docs (#60)


SBlokhuizen:
In #61 and #58, we have ParaView state files which I would like to tag with a specific version of IMAS-ParaView for which they work. As these depend on the fix in #57, I propose we release a new patch  ``v2.3.1`` which we can tag these state files with. What do you think @maarten-ic?

maarten-ic:
Two main questions:
1. Wouldn't you want the updated state files be part of this release? Otherwise the released documentation contains outdated state files?
2. Why increment the patch number only? The included commits indicate that there's new features included as well, so following semver shouldn't the version become 2.4.0?

SBlokhuizen:
As discussed, let's create a new release 2.4.0 in a separate PR, and update the linked state files to point to version 2.4.0.

PR #63, opened 2026-04-14
Add version tags for new / changed plugins

SBlokhuizen:
- [x] Add tag to new Camera Reader

### DIFF
--- a/docs/source/plugins/1d_profiles_mapper.rst
+++ b/docs/source/plugins/1d_profiles_mapper.rst
@@ -3,7 +3,8 @@
 1D Profiles Mapper
 ==================
 
-.. versionchanged:: 2.4.0
+.. versionadded:: 2.0.0
+.. versionchanged:: 2.4.0 Name of psi array is now auto-detected.
 
 The **1D Profiles Mapper** is a ParaView filter designed to map 1D profiles onto a 2D flux grid. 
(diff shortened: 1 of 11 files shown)


docs/source/plugins/1d_profiles_mapper.rst
```diff
@@ -3,6 +3,8 @@
 1D Profiles Mapper
 ==================
 
+.. versionchanged:: 2.4.0
```
maarten-ic: Can you indicate in one sentence what has changed in this version?

Is it useful to have a `.. versionadded::` for these as well? This is e.g. what the Python docs do ([example](https://docs.python.org/3/library/typing.html#typing.TypeVarTuple.has_default)).

docs/source/plugins/los_reader.rst
```diff
@@ -3,6 +3,8 @@
 Line of Sight Reader
 ====================
 
+.. versionchanged:: 2.3.0
```
maarten-ic: Idem

SBlokhuizen: I have added a short changelog for each `versionchanged` tag. I have also added additional changes down to the first public release of IMAS-ParaView (v2.0.0).

maarten-ic (approved):
LGTM!

PR #65, opened 2026-04-21
Beautify landing page

SBlokhuizen:
This changes the docs landing page to be a bit nicer.

### DIFF
--- a/docs/source/_static/gallery-shuffle.js
+++ b/docs/source/_static/gallery-shuffle.js
@@ -2,12 +2,12 @@
 document.addEventListener("DOMContentLoaded", () => {
     const source = document.getElementById('all-gallery-source');
     const cards = Array.from(source.querySelectorAll('.sd-col'));
-    for (let i = cards.length - 1; i > 0; i--) {
-        const j = Math.floor(Math.random() * (i + 1));
+    for (let i = 0; i < 3; i++) {
+        const j = i + Math.floor(Math.random() * (cards.length - i));
         [cards[i], cards[j]] = [cards[j], cards[i]];
     }
+    const row = source.querySelector('.sd-row');
     cards.slice(3).forEach(card => card.remove());
-    cards.slice(0, 3).forEach(card => source.querySelector('.sd-row').appendChild(card));
+    cards.slice(0, 3).forEach(card => row.appendChild(card));
     source.style.display = '';
-    source.removeAttribute('id');
 });
--- a/docs/source/_static/gallery-shuffle.js
+++ b/docs/source/_static/gallery-shuffle.js
@@ -1,0 +1,13 @@
+// displays a random selection of 3 gallery cards
+document.addEventListener("DOMContentLoaded", () => {
+    const source = document.getElementById('all-gallery-source');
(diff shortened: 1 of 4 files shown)


docs/source/index.rst
```diff
@@ -1,41 +1,111 @@
...
+
+.. grid:: 1 2 3 3
+   :gutter: 3
+   :margin: 2 2 2 2
+
+   .. grid-item-card::
+      :link: installing
+      :link-type: ref
+
+      .. raw:: html
```
maarten-ic: Do you require raw HTML here?

Panel [only uses sphinx plugins](https://github.com/holoviz/panel/blob/main/doc/index.md#learn-panel) to achieve these cards (in myst markdown, but should be possible in rst as well)

If it's just for the icons, you could have a look at the theme documentation (https://sphinx-immaterial.readthedocs.io/en/latest/inline_icons.html) as well

maarten-ic: The "Installation" text moves a bit weirdly on mouse-over (it moves slightly down, while the icon next to it moves slightly up), but maybe that's because of the custom HTML?

docs/source/_static/gallery-shuffle.js
```diff
@@ -0,0 +1,13 @@
...
+    const source = document.getElementById('all-gallery-source');
+    const cards = Array.from(source.querySelectorAll('.sd-col'));
+    for (let i = cards.length - 1; i > 0; i--) {
+        const j = Math.floor(Math.random() * (i + 1));
+        [cards[i], cards[j]] = [cards[j], cards[i]];
+    }
+    cards.slice(3).forEach(card => card.remove());
+    cards.slice(0, 3).forEach(card => source.querySelector('.sd-row').appendChild(card));
+    source.style.display = '';
+    source.removeAttribute('id');
```
maarten-ic: What's the purpose of that?

docs/source/_static/gallery-shuffle.js
```diff
@@ -0,0 +1,13 @@
+// displays a random selection of 3 gallery cards
+document.addEventListener("DOMContentLoaded", () => {
+    const source = document.getElementById('all-gallery-source');
+    const cards = Array.from(source.querySelectorAll('.sd-col'));
+    for (let i = cards.length - 1; i > 0; i--) {
+        const j = Math.floor(Math.random() * (i + 1));
+        [cards[i], cards[j]] = [cards[j], cards[i]];
+    }
```
maarten-ic: You could shuffle only the first three items (since the rest is discarded anyway)? But should be fast anyway so doesn't matter much

maarten-ic (commented):
Looks nice! But could we get rid of the custom HTML?

docs/source/index.rst
SBlokhuizen: I cannot replicate the mouse-over behaviour you describe, for me it slightly enlarges the text on mouse hover. Perhaps it's a scaling issue? 

I have replaced the icons by the inline icons you mentioned, could see if it looks better now?

maarten-ic (approved):
Nice!

PR #67, opened 2026-04-22
add support for converting 2D structured GGD grids

### LINKED ISSUE #66: errors reading plasma_profiles
Hello,

I am getting an error while reading the GGD from plasma_profiles.
I am using IMAS-ParaView/2.3.0-foss-2023b
and the Data Dictionary 4.1.1

The uri loads correctly
I can see the plasma_profiles in the list of supported IDS. 

The problem might well be on my side (writing the GGD incorrectly).
Can you confirm that the reader has been tested for the plasma_profiles?

I am writing a simple (R,z) grid and putting the electron density and temperature and the ion density and temperature on the GGD, using the (R,z) representation.

The GGD_grid has been filled as well (perhaps also incorrectly?)

Here is my uri on the SDCC:

imas:hdf5?path=/home/ITER/polif/test_bridge/test_interferometer/

If you confirm that the reader has been tested on the plasma_profiles IDS, I will dig deeper in my script that writes into the GGD.

Thank you

SBlokhuizen:
Add conversion of structured GGD grids using the GGD reader, such as the `plasma_profiles` IDS in following URI:

```
imas:hdf5?path=/home/ITER/polif/test_bridge/test_interferometer/
```

This was also able to succesfully visualize a structured GGD grid from the [GGD Fortran library examples](https://github.com/iterorganization/GGD/blob/develop/examples/f90/ids_grid_example1_2dstructured_manual.f90).


Yannicked (approved):

docs/source/cli.rst
```diff
@@ -154,6 +159,9 @@ The ``vtk2ggd`` converter has specific requirements for the VTK input files it c
 * When providing ``.vtpc`` files, each ``vtkPartitionedDataSet`` should contain **exactly one partition** containing a ``vtkUnstructuredGrid``, otherwise the 
   partition is skipped. Each ``vtkUnstructuredGrid`` will be converted into a separate grid subset in the GGD.
 
+.. note::
+   It is currently not possible to convert ``vtkStructuredGrid`` s to GGD grids.
```
maarten-ic: I believe you can escape the space like this to avoid the hanging 's'

```suggestion
   It is currently not possible to convert ``vtkStructuredGrid``\ s to GGD grids.
```

docs/source/cli.rst
```diff
@@ -34,6 +34,11 @@ The vtpc file contains a vtkPartitionedCollection, which can directly be loaded
 for visualization. The vtu files contain a vtkUnstructuredGrid, and can also be loaded into
 Paraview directly.
 
+.. note::
+   Both unstructured and structured GGD grids are supported for conversion to VTK.
```
maarten-ic: Only 2D structured grids?

imas_paraview/io/read_geom_structured.py
```diff
@@ -0,0 +1,125 @@
...
+
+def _get_coord_type(space) -> int:
+    """Return the coordinate identifier for a space."""
+    coordinates_type = space.coordinates_type
+    if isinstance(coordinates_type[0], IDSStructure):
+        return int(coordinates_type[0].index)
+    return int(coordinates_type[0])
+
+
+def _extract_1d_coords(space) -> np.ndarray:
```
maarten-ic, line 115: Check that the space has only one dimension?

The GGD could contain a 3D layered grid, where one space consists of (unstructured) R,Z points, and the other space has an array of phi values. That's not supported by this reader.

imas_paraview/util.py
```diff
@@ -477,3 +477,33 @@ def ensure_unique_name(name, existing_names):
         counter += 1
         unique_name = f"{name} #{counter}"
     return unique_name
+
+
+def is_structured_grid(grid_ggd):
+    """Return True if grid_ggd is a structured grid.
```
maarten-ic: Perhaps be more precise that you're only looking for 2D structured grids?

3D structured grids and layered grids are not supported, right?

maarten-ic (commented):
Code looks good, but can you be a bit more precise on the supported (and unsupported) use cases?

imas_paraview/io/read_geom_structured.py
SBlokhuizen: The 3D layered grid you describe should be guarded against since we do check that the [length of the geometry is equal to 1](https://github.com/iterorganization/IMAS-ParaView/pull/67/changes#diff-4f4ffd0635f0159ce7a8f86ece2874d8e712862a18dad443a005170339717d1dR506). 

I did have a check before to check the space only has a single dimension, but I found that the GGD from the [fortran examples](https://github.com/iterorganization/GGD/blob/develop/examples/f90/ids_grid_example1_2dstructured_manual.f90) actually defines both the points and edges of the 1D spaces. I am not sure how common this is, but I wanted to make sure the reader was compatible with this example case. In the reader we now simply ignore any entries in the `objects_per_dimension` AoS besides the first one. 

We could log a warning if the space has more than one dimension instead?

maarten-ic: > We could log a warning if the space has more than one dimension instead?

Would be good to add indeed, if only as sanity check

maarten-ic (approved):

PR #71, opened 2026-04-24
add tracer example to gallery

SBlokhuizen:


paulotex (approved):
I think this is what we are asking for.

docs/source/gallery/examples/distributions_tracing/description.yaml
```diff
@@ -0,0 +1,7 @@
+title: Distributions Markers Particle Tracing
+author: S. Blokhuizen
+description: |
+  Visualization of 10 synthetic electron marker particles tracing helical orbits
+  over time. The dataset is synthetic and constructed for demonstration and testing purposes.
+uri: imas:hdf5?path=/home/ITER/blokhus/public/synthetic_distribution
```
maarten-ic, line 6: Since you constructed this data yourself, would it be possible to publish it openly on the internet as netCDF?

maarten-ic (approved):
LGTM!

docs/source/gallery/examples/distributions_tracing/description.yaml
SBlokhuizen: Good idea, I have uploaded this synethetic distributions IDS as NetCDF to [Zenodo](https://zenodo.org/records/19879666).  I have also added an optional `link` entry to the description.yaml, so people can provide external links to their dataset. I have kept the URI there so people can still use the state file when they are on SDCC.

maarten-ic (approved):
Nice!

PR #73, opened 2026-04-28
Ensure CI is run when a PR from a fork is created

SBlokhuizen:


maarten-ic (approved):

.github/workflows/publish_pypi.yml
```diff
@@ -2,6 +2,7 @@ name: build-wheel-and-publish-pypi
 
 on:
   push:
+    branches: [main, develop]
```
SimonPinches, line 5: Do we really want to publish all changes to the develop branch to PyPi ?

SBlokhuizen: Hi Simon, 

This change ensures that the build job runs in CI (on `main`, `develop`, and pull requests) to verify that the package builds correctly. It will only publish to PyPi when a tag is pushed (as you can see in [L32](https://github.com/iterorganization/IMAS-ParaView/blob/develop/.github/workflows/publish_pypi.yml#L32)).

PR #72, opened 2026-04-28
Added auto-detect for the Grid Psi array for the 1D profile. 

sqkorving:
Core_sources have structure 'Source (Type) Grid Psi'

PR #74, opened 2026-04-29
Dynamically find psi grid in 1d profiles mapper

SBlokhuizen:


imas_paraview/plugins/profiles_1d_mapper.py
```diff
@@ -72,7 +72,21 @@ def RequestData(self, request, inInfoVec, outInfo):
             return 1
         psi = input0.PointData[psi_array_name]
         psi_grid = psi.GetArrays()[0]
-        psi_profiles = input1.RowData["Grid Psi"]
+
+        # Auto-detect the Psi grid
```
maarten-ic: Might be nice to briefly mention the use case here (the core_sources scenario from @sqkorving from #72 )?

maarten-ic (approved):

PR #75, opened 2026-05-06
Magnetics reader

SBlokhuizen:
Introduces a new reader to visualize different kinds of geometries from the [magnetics](https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/magnetics.html) IDS, namely:
- Flux loops
- B-field probes (poloidal and toroidal)
- Rogowski coils


Yannicked (approved):

docs/source/plugins/magnetics_reader.rst
```diff
@@ -0,0 +1,59 @@
...
+Refer to the :ref:`using the GGD Reader` for detailed instructions on:
+
+- :ref:`Loading an URI <loading-an-uri>`: How to provide the file path or select a dataset.
+- :ref:`Loading an IDS <loading-an-ids>`: How to load a dataset and display the grid.
+- :ref:`Selecting attribute arrays <selecting-ggd-arrays>`: How to choose and visualize attributes.
+
+Visualized Elements
+-------------------
+
+Flux loops and Rogowski coils are visualized as closed polylines representing the wire geometry.
```
maarten-ic, line 42: Perhaps add a note here that the displayed flux loop geometry is a linear interpolation between the points defined in the IDS (which may appear to cross the plasma, while in reality the loop would run inside the wall).

See also the SVG linked from https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/magnetics.html#magnetics-flux_loop-position ("Click here for further documentation.")

imas_paraview/plugins/magnetics.py
```diff
@@ -0,0 +1,238 @@
...
+        self._populate_b_field_probes()
+        self._populate_rogowski_coils()
+
+        self._selectable = list(self.selectable_map.keys())
+
+    def _populate_flux_loops(self):
+        """Populate selectable_map with flux loops."""
+        for i, loop in enumerate(self._ids.flux_loop):
+            name = self._create_name("Flux loop", loop, i)
+            if len(loop.position) < 2:
```
maarten-ic: Would it be difficult to support the special case for full flux loops? From DD docs:

> For simplicity, full flux loops can be described by one single position

imas_paraview/plugins/magnetics.py
```diff
@@ -0,0 +1,238 @@
...
+
+        Args:
+            name: Name of the probe.
+            probe: B-field probe IDS node.
+            kind: Either "pol" or "phi" for poloidal or toroidal probe.
+        """
+        if not (
+            probe.position.r.has_value
+            and probe.position.phi.has_value
+            and probe.position.z.has_value
```
maarten-ic: Didn't we have a utility function for this?

imas_paraview/plugins/magnetics.py
```diff
@@ -0,0 +1,238 @@
...
+        """
+        for block_id, name in enumerate(self._selected):
+            selected = self.selectable_map[name]
+
+            if isinstance(selected, (FluxLoop, RogowskiCoil)):
+                vtk_geom = self._create_loop(selected)
+            else:  # Bfield probe
+                vtk_geom = self._create_b_field_probe_arrow(selected)
+
+            if vtk_geom is None:
```
maarten-ic: Neither function should return None, right?

maarten-ic (commented):
Looks good! Couple of comments / questions. Let me know if anything's unclear!

imas_paraview/plugins/magnetics.py
maarten-ic: Double-checking: how is a full flux loop (such as the ones for WEST) visualized now? As a single point or do you plot a circle at `(r, z, phi=0...2pi)`?

SBlokhuizen: A flux loop containing a single point is indeed just visualized as a single point. If the user wants to visualize it as a full circle, they can apply a rotational extrusion filter to it. Would you rather we do this extrusion in the plugin itself?

maarten-ic: > Would you rather we do this extrusion in the plugin itself?

If there's a mix of full flux loops and saddle loops loaded by the plugin, then you don't want to extrude them all (which is what a filter after the plugin would do). I don't know how relevant this scenario is though.

Let's check with IO?

SBlokhuizen: > If there's a mix of full flux loops and saddle loops loaded by the plugin, then you don't want to extrude them all (which is what a filter after the plugin would do).

Indeed, users would then first need to extract a subset of loops using e.g. the `ExtractBlock` filter before applying the extrusion. 

Hi @paulotex, do you have a preference for either one of the two methods?

paulotex: I prefer minimal action from the user. What about, we star t with what we have now, flux loop with a single point gets that single point displayed. But we add a checkbox (extrude) that when checked extrudes the single point loops 360 degrees. I am assuming they are extruded toroidally along phi.

paulotex: I can't find a short label name for that checkbox that means "Automatically Extrude Single Point Flux Loops", but this is what I meant when I suggested the label "Extrude": it is implicit it only cats on the single point loops.

If it is simple to attach such a check box to each single point flux loop (perhaps on the array table?) than we can do that, but I think a single check box under that table that applies to all single point loops is ok.

image

SBlokhuizen: Hi @paulotex, I have implemented a checkbox in the plugin settings named ``Extrude Full Flux Loops``, with additional information when you hover over it. When enabled, it will automatically extrude any flux loop that only contains a single point. I have also added this information to the documentation page for the reader. Please let me know if this works for you!

imas_paraview/plugins/magnetics.py
```diff
@@ -55,11 +58,16 @@ class BFieldProbe:
 @smproxy.source(label="Magnetics Reader")
 @smhint.xml("""<ShowInMenu category="IMAS Tools" />""")
 class MagneticsReader(GGDVTKPluginBase):
+    _FULL_LOOP_N_POINTS = 100
```
maarten-ic: Should this be a setting?

maarten-ic (approved):

SimonPinches:
I know I'm late to comment, but personally I think I'd always extrude in the toroidal direction since that's what the loop looks like.  We don't need Paraview and all it's 3D capabilities to plot a single point in the poloidal plane.

paulotex (approved):

SBlokhuizen:
@SimonPinches, I agree. I have discussed this with @paulotex and we decided to enable the rotational extrusion for the single-point flux loops by default.

PR #81, opened 2026-06-15
Fix missing position issue

### LINKED ISSUE #80: jacobi error in paraview (position reader and WEST data)
Please try the following URI with the position reader:
`imas:hdf5?path=/home/ITER/tribolp/public/imasdb/WEST/57410/0`

When trying to display any of the available IDS (magnetics, barometry, langmuir_probes) I get the following error in paraview:
```
:Jacobi: Error extracting eigenfunctions
Generic Warning: In vtkMath.cxx, line 778
vtkMath::Jacobi: Error extracting eigenfunctions
```

Can you please check what might be happening here?

SBlokhuizen:
Fixed the position reader to skip devices with unfilled positions. Also allow visualizing positions in cases where `phi` is left empty, for example in the case of 2D axisymmetric geometries, such as the WEST example in #80.

### DIFF
--- a/imas_paraview/ids_util.py
+++ b/imas_paraview/ids_util.py
@@ -177,9 +177,15 @@ def cart_vector_has_value(vec):
         return False
 
 
-def cyl_vector_has_value(vec):
-    """Check if a lazy-loaded cylindrical IMAS vector quantity has data."""
+def cyl_vector_has_value(vec, check_phi=True):
+    """Check if a lazy-loaded cylindrical IMAS vector quantity has data.
+
+    Args:
+        vec: IMAS structure with r, phi, z fields.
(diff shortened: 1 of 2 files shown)


imas_paraview/ids_util.py
```diff
@@ -177,9 +177,15 @@ def cart_vector_has_value(vec):
         return False
 
 
-def cyl_vector_has_value(vec):
-    """Check if a lazy-loaded cylindrical IMAS vector quantity has data."""
+def cyl_vector_has_value(vec, check_phi=True):
```
maarten-ic: I'd prefer another helper `rz_vector_has_value()`: the `check_phi` argument feels a bit awkward to me

maarten-ic (approved):

PR #82, opened 2026-06-16
Release 2.4.0

SBlokhuizen:
# Release notes IMAS-ParaView 2.4.0

 This release introduces four new ParaView readers and includes several improvements and bug fixes.

### 4 new ParaView reader plugins
- [(Shattered) Pellets Reader](https://imas-paraview.readthedocs.io/en/stable/plugins/pellets.html) - for visualizing shattered pellet injector fragment positions and velocities from the `spi` IDS
- [Camera Reader](https://imas-paraview.readthedocs.io/en/stable/plugins/camera_reader.html) - for visualizing camera sightlines from the `camera_ir` and `camera_visible` IDSs, including the option to match the ParaView camera to a selected camera for a point-of-view visualization
- [Distributions Reader](https://imas-paraview.readthedocs.io/en/stable/plugins/distributions_reader.html) - for visualizing particle distribution marker data from the `distributions` IDS
- [Magnetics Reader](https://imas-paraview.readthedocs.io/en/stable/plugins/magnetics_reader.html) - for visualizing flux loops, B-field probes, and Rogowski coils from the `magnetics` IDS

### GGD reader improvements
- [Added support](https://github.com/iterorganization/IMAS-ParaView/pull/67) for loading structured 2D GGD grids in the GGD reader
- Added option for [cylindrical to Cartesian coordinate transformation](https://imas-paraview.readthedocs.io/en/stable/plugins/ggd_reader.html#converting-vector-components) of GGD vector components

### Documentation improvements
- Overhauled [Gallery section](https://imas-paraview.readthedocs.io/en/stable/gallery/gallery.html) to make it easier for users to add new examples. 
- Added several examples showing example workflows to the [Gallery section](https://imas-paraview.readthedocs.io/en/stable/gallery/gallery.html).
-  Redesigned [Landing page](https://imas-paraview.readthedocs.io/en/stable/index.html) with examples and links to the most important pages 
- [Plugin](https://imas-paraview.readthedocs.io/en/stable/plugins.html) documentation page now features image previews with examples, making it much easier to navigate the different plugins.
- Added version tags to plugin documentation indicating when each plugin was introduced or changed.

### Bug fixes
- Fixed psi grid and array loading in the 1D profiles mapper (#57, #74)
- Fixed an issue where readers would fail to load an IDS with occurrence > 0 if occurrence 0 was missing (#79)
- Fixed the position reader to skip devices with unfilled positions, and allow visualization where phi is empty (#81)

PR #86, opened 2026-06-25
Clean-up documentation

SBlokhuizen:
Updates the documentation according to the comments received by @olivhoenen on #82.


olivhoenen:
Thanks @SBlokhuizen, two remaining comments on the documentation:
- there are few mentions still to git.iter.org and sharepoint (in installing.rst and asv.conf.json), they should be replaced with their github/rtd counterparts
- w.r.t pip install from PyPI, would it make sense to explicity have that route as one example in installing.rst? could even give a command like ```PV_PLUGIN_PATH=`python -c "import imas_paraview; print(imas_paraview.__path__[0])"` ```

Yannicked:
> * there are few mentions still to git.iter.org and sharepoint (in installing.rst and asv.conf.json), they should be replaced with their github/rtd counterparts

The references to git.iter.org and sharepoint are removed in #87.

DaanVanVugt (approved):

DaanVanVugt (approved):

PR #92, opened 2026-08-05
Distributions fix

paulotex:
captures an exception in the distributions IDS that would be triggered with some species did not have a name defined. This is particular relevant for ASTRA scenarios.

### DIFF
--- a/imas_paraview/plugins/distributions.py
+++ b/imas_paraview/plugins/distributions.py
@@ -113,7 +113,10 @@ class DistributionsReader(GGDVTKPluginBase, is_time_dependent=True):
         coord_names_seen = {self._NONE_LABEL}
 
         for i, dist in enumerate(self._ids.distribution):
-            dist_name = self._create_dist_name(dist) or f"distribution {i}"
+            try:
+                dist_name = self._create_dist_name(dist)
+            except AttributeError:
+                dist_name = f"distribution {i}"
 
             if len(dist.markers) == 0:
                 logger.warning("'%s' does not contain any markers, skipping", dist_name)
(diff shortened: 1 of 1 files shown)


paulotex:
@SBlokhuizen I am aware you contract is cloed, but could you have a look if this seems ok to you?
Astra scenarios don't fill the 'name' of most distribution species (ions, neutrals). Instead of an exception and a stack trace, I use the species index as a name.

maarten-ic:
Hi Paulo, an `AttributeError` should only be possible when you load data in a DD version that doesn't define any of the data fields the logic is looking for. Maybe it's related to the rename from `label` -> `name`? https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/distributions.html#distributions-distribution-species-ion-name

It'd be better to handle that case in the _create_dist_name function instead of wrapping it in a try-except at the call side IMO

paulotex:
Hi @maarten-ic thanks for the quick update.
Indeed, this is data with DD 3.38, which has `label` but not `name`.
Why do you think it is better to handle this in `_create_dist_name`? 
I prefer to not start adding code to look for `label` instead of `name`, that is, I prefer to not add code that tries to handle old DD versions.
I was hoping that with this minimal "fix", old DD versions could still be read, although losing some of the metadata (the `label` in this case).

paulotex:
@maarten-ic 
Ok, I just added some lines in `_create_dis_name` that tries `name` and then `label`. Does this look better to you? Would you do it differently perhaps?

maarten-ic:
Hi Paulo,

> Why do you think it is better to handle this in `_create_dist_name`?

Because the method could (though it's probably not right now) be used from multiple places, therefore it's better to handle the exception inside the method. Regardless of the solution 😉 

> I prefer to not start adding code to look for `label` instead of `name`, that is, I prefer to not add code that tries to handle old DD versions. I was hoping that with this minimal "fix", old DD versions could still be read, although losing some of the metadata (the `label` in this case).

IMAS-ParaView already does this in multiple places, so for consistency I'd handle it here as well. See, for example, https://github.com/iterorganization/IMAS-ParaView/blob/2fba6cd4c1524e849b7578f066ada7bc9dbb2c1c/imas_paraview/ids_util.py#L144-L150

(Thanks @SBlokhuizen )

maarten-ic:
> @maarten-ic Ok, I just added some lines in `_create_dis_name` that tries `name` and then `label`. Does this look better to you? Would you do it differently perhaps?

Looks fine to me!

SBlokhuizen:
Hi Paulo, the changes look fine to me! The CI fails because there is an linting/formatting issue, you can check the issue locally by running the following:
```
ruff format imas_paraview       # Format code
ruff check imas_paraview        # Lint code
```

You can read more about this [in the documentation](https://imas-paraview.readthedocs.io/en/stable/code_style.html). If you need any help, please let me know!

paulotex:
Thanks, this has been very helpful. I made some minor changes to make it more readable. I removed the `try/except` and used `hasattr`, and used a helper function to select either DD4 name or DD3 label. If you agree this looks ok (and ruff and lint are happy) I'll merge.

maarten-ic:
Thanks Paulo! This looks ok to me, though `ruff format` has some complaints. Can you run it and push the changes (see [Sebbe's post above](https://github.com/iterorganization/IMAS-ParaView/pull/92#issuecomment-5192650272))? Feel free to send one of us a message on Slack if you need any help!

SBlokhuizen (approved):

maarten-ic (approved):

